PR #59, opened 2025-08-13
fix(conversion): fixed using wrong timestamp

Yukiniro:
The original logic `lastCanvasEndTimestamp = timestamp + duration;` uses the timestamp of the original frame, but `padFrames` calculates the difference between it and `lastCanvasTimestamp`. This will result in an incorrect end time. For example, when trimming a video to [10, 20], the final video will be 20s instead of 10s.

### DIFF
--- a/src/conversion.ts
+++ b/src/conversion.ts
@@ -757,7 +757,7 @@ export class Conversion {
 						}
 
 						let adjustedSampleTimestamp = Math.max(sample.timestamp - this._startTimestamp, 0);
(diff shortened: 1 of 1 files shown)


Vanilagy:
Great catch, can you fix it for the other case too, analogously? (the VideoSample iteration a bit further down)

Yukiniro:
OK, I'll give it a try

Vanilagy (approved):
Perfect, thanks!

Vanilagy:
deployed in 1.7.5

PR #60, opened 2025-08-14
Fix UrlSource single-byte buffer on Range Request HTTP 200

devPablo:
## Issue
When loading a video though the `UrlSource` Source, the following error would appear: `Input has an unsupported or unrecognizable format.`

## Solution
Unsure on where exactly the error was first introduced, in `source.js`, in function `_retrieveSize `of `UrlSource`, when performing a successful **Range Request** (HTTP 200), the header `Content-Range` buffer size would equate to **1**, fix was to skip it and force a full GET request.

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -316,6 +316,8 @@ export class UrlSource extends Source {
 			this._fullData = await rangeResponse.arrayBuffer();
 			if (this._fullData.byteLength !== 1) {
 				return this._fullData.byteLength;
+			} else {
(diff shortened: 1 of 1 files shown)


Vanilagy (changes requested):
Awesome! I'd add an `else` block with just a comment inside, that reads something like "The server responded with 200 but returned only the requested range, so skip the response", so it's clear why the check happens.

Vanilagy (approved):

Issue #70, opened 2025-08-20
[BUG] Fragments circular linked cause infinite while loop in isobmff-demuxer performFragmentedLookup


hikariNTU:
**Bug description**:

When decoding mp4 from cropped youtube video, reading duration will cause infinite while loop and stuck the browser tab.

**File**:
https://drive.google.com/file/d/1jFbm4eH4BV2FF6OusVap_bFxo_-Sh2eL/view?usp=sharing

**Reproduce**:

```ts
  const input = new Input({
    formats: ALL_FORMATS,
    source: new BlobSource(file), // an mp4
  });
  const duration = await input.computeDuration(); // stuck in while loops
```

**Causing (debugger)**:

```ts
while (metadataReader.pos < sourceSize) {
	if (prevFragment) {
		const trackData = prevFragment.trackData.get(this.internalTrack.id);
		if (trackData && trackData.startTimestamp > latestTimestamp) {
			break;
		}

		if (prevFragment.nextFragment) {
			metadataReader.pos = prevFragment.nextFragment.moofOffset + prevFragment.nextFragment.moofSize;
			prevFragment = prevFragment.nextFragment; // [BUG] circular linked
			continue;
		}
	}
```
https://github.com/Vanilagy/mediabunny/blob/main/src/isobmff/isobmff-demuxer.ts#L2385

<img width="1203" height="279" alt="Image" src="https://github.com/user-attachments/assets/ef5a1484-9cec-4a5d-90fa-88d909fcf888" />

---

**Potential Fix**:

```diff
if (prevFragment.nextFragment) {
	// Skip ahead quickly without needing to read the file again
	metadataReader.pos = prevFragment.nextFragment.moofOffset + prevFragment.nextFragment.moofSize;
+	if (prevFragment.nextFragment === prevFragment) {
+		// We're at the last fragment, no need to keep searching
+		break;
+	}
	prevFragment = prevFragment.nextFragment;
	continue;
}
```

**File Info (ffprobe)**:

```text
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '1080-30min.mp4':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomiso6iso2avc1mp41
    title           : foobar
    artist          : foobar
    encoder         : Lavf60.3.100
  Duration: 00:27:00.31, start: 0.000000, bitrate: 1747 kb/s
[...]

PR #71, opened 2025-08-20
fix: infinite loop in IsobmffTrackBacking when processing fragments

hikariNTU:
Fix:
- #70 

---

TBH I am more curious about how does `nextFragment` become self linked in the first place


hikariNTU:
Oh, there's like 4 `currentFragment = currentFragment.nextFragment` in the code

I think we should fix the nextFragment issue directly

PR #74, opened 2025-08-25
Add descriptive metadata read/write support

Vanilagy:
This PR adds read/write support for *descriptive metadata* for all container formats. Descriptive metadata is stuff like title, artist, album, cover art, release date, etc.


src/metadata.ts
```diff
@@ -0,0 +1,85 @@
+// TODO: Give this a different name. This clashes with a web concept. Question is, what?
```
hobodrifterdavid, line 1: https://developer.mozilla.org/en-US/docs/Web/API/Media_Session_API

PR #86, opened 2025-09-02
Do not make UrlSource fail if the size of the file is smaller than 512kb (+ add a Vitest browser mode test)

JonnyBurger:
A 64kb test file loaded with `UrlSource` failed with

```
Needed at least 524288 bytes, got only 67304.
```

The first commit fixes this bug.
The remaining commits add a test, that actually executes in the browser using Vitest browser mode.

This allows us to write more tests in the future that rely on the WebCodecs API.


Vanilagy (changes requested):
Just a couple of nitpicks. The UrlSource fix should be correct, the other place `URL_SOURCE_MIN_LOAD_AMOUNT` is used is automatically clamped elsewhere, so that's fine.

Also, I'm more of a fan of lesser, larger files, instead of one file per test. But, let's leave it for now until we do the proper test pass.

JonnyBurger:
The tests need to be separated into 2 files: 

I set up one test suite which runs in Node (faster) and one that runs in the browser (has WebCodecs and a HTTP server to simulate URLs)

Applied your suggestions!

Vanilagy (approved):

PR #87, opened 2025-09-02
Enable ESLint as formatter

JonnyBurger:
I was working on a different computer today, and the formatting on save did not work out of the box for me.

I had to manually enable this setting, most people have it probably but explicitly enabling it for this project makes it easier for contributors.


JonnyBurger:
Also fine, works as well!

Vanilagy (approved):
Dope

PR #88, opened 2025-09-02
Fix tfdt logic

cr3314:
By definition, Track Fragment Base Media "Decode" Time.

It is currently problematic with convert to fmp4.
1. `MediaSource`, `SourceBuffer.mode = 'segments'`: When there is b-frame, using PTS (e.g. 24fps, 2 b-frame, 0.08333) could result in `buffered` range gap that `<video>` get stuck.
2. The result fmp4 file duration would also be longer by an extra first PTS amount.

### DIFF
--- a/src/isobmff/isobmff-boxes.ts
+++ b/src/isobmff/isobmff-boxes.ts
@@ -1208,7 +1208,7 @@ export const tfdt = (trackData: IsobmffTrackData) => {
 	assert(trackData.currentChunk);
 
 	return fullBox('tfdt', 1, 0, [
-		u64(intoTimescale(trackData.currentChunk.startTimestamp, trackData.timescale)), // Base Media Decode Time
+		u64(intoTimescale(trackData.currentChunk.startDecodeTimestamp, trackData.timescale)), // Base Media Decode Time
 	]);
 };
 
--- a/src/isobmff/isobmff-muxer.ts
+++ b/src/isobmff/isobmff-muxer.ts
@@ -50,6 +50,7 @@ export type Sample = {
 type Chunk = {
 	/** The lowest presentation timestamp in this chunk */
 	startTimestamp: number;
+	startDecodeTimestamp: number;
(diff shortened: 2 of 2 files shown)


ghost:
Apparently, the result is the same as https://github.com/Vanilagy/mediabunny/pull/88/files#diff-cfdf660642e5b92ba0fb3dc2d0df213e985b858a5a2f757cd8d8004a6b1d1986R850
I am tripping over a bug that was in the old `mp4-muxer` when migrating. Sorry for that.

Vanilagy:
All good! Is the `<video>` playback issue fixed for you now, though?

ghost:
> All good! Is the `<video>` playback issue fixed for you now, though?

Yes. Thanks!

Vanilagy:
Awesome. thanks for the PR anyway, I appreciate all help

ghost:
After some more reading and trying, I learnt that the current behavior of mediabunny for b-frame would be `ffmpeg` with `-movflags negative_cts_offsets`.
When trying to mix with fmp4 from `ffmpeg` default setting, MSE (at least on chrome) would have a playback gap problem (DTS is discontinuous). Anyways, the situation can be avoided or worked around by always reconverting the file with Mediabunny first. 👍

Vanilagy:
Yes, Mediabunny uses negative CTS offsets quite frequently! Files support it, so it makes use of it, but I understand that it can cause discontinuities when mixed with other stuff. Do you think I should add some info my docs anywhere regarding this?

ghost:
I am not an expert and I could be totally wrong with what I say. This is probably not something someone would normally hit.
So, I wouldn't risk adding to doc. 😄

And, although useful for some cases, if we keep adding support for FFmpeg behaviour for things like this, eventually this library will require the same level of maintenance as FFmpeg.

Vanilagy:
Yeah exactly, and I don't wanna go there. Will be hard enough to maintain as-is.

Issue #90, opened 2025-09-03
HDR input file + forceTranscode?


hamzaaftab:
Thanks for putting together a fantastic library. Running into an issue with a specific file not being able to get converted (suspect it has to do with HDR colormap).

Environment:
- Mac OS
- Chrome Version 139.0.7258.155 (Official Build) (arm64)

Code:
```
const input = new Input({
  source: new BlobSource(file), // (see ffprobe output below for what the file is)
  formats: ALL_FORMATS,
});
const output = new Output({ format: outputFormat, target });
const conversion = await Conversion.init({
  input,
  output,
  video: {
    forceTranscode: true,
  },
});
```

Error:

```
Encoding error.

```

Interestingly when I try to use media sink + media output and iterate over all the samples to convert it that way, I can get the samples from original file, but then writing back still fails with same error.

Crucially: Using Canvas to convert each input sample from HDR to SDR does fix the issue, so perhaps it is related to HDR color profile. Currently trying that approach on top of sink + encode loop, but open to other ideas too.

ffprobe output on input file:
```
ffprobe version 7.1.1 Copyright (c) 2007-2025 the FFmpeg developers
  built with Apple clang version 16.0.0 (clang-1600.0.26.6)
  configuration: --prefix=/opt/homebrew/Cellar/ffmpeg/7.1.1_1 --enable-shared --enable-pthreads --enable-version3 --cc=clang --host-cflags= --host-ldflags='-Wl,-ld_classic' --enable-ffplay --enable-gnutls --enable-gpl --enable-libaom --enable-libaribb24 --enable-libbluray --enable-libdav1d --enable-libharfbuzz --enable-libjxl --enable-libmp3lame --enable-libopus --enable-librav1e --enable-librist --enable-librubberband --enable-libsnappy --enable-libsrt --enable-libssh --enable-libsvtav1 --enable-libtesseract --enable-libtheora --enable-libvidstab --enable-libvmaf --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libxvid --enable-lzma --enable-libfontconfig --enable-libfreetype --enable-frei0r --enable-libass --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-libspeex --enable-libsoxr --enable-libzmq --enable-libzimg --disable-libjack --disable-indev=jack --enable-videotoolbox --enable-audiotoolbox --enable-neon
[...]

PR #91, opened 2025-09-03
Allow forcing rerender (helps with some HDR inputs)

hamzaaftab:


Vanilagy:
Closed, see https://github.com/Vanilagy/mediabunny/issues/90#issuecomment-3252665344

Issue #55, opened 2025-08-11
Webm Transparency


shealan:
Such an incredible library, thank you! <3

I would love to see transparent video support. Are there any plans to enable it for webm videos, as the VP9 codec does support it.

Some relevant info here: https://rotato.app/blog/transparent-videos-for-the-web

My initial testing showed that it's not currently possible. The alpha area is always rendered as black.

Vanilagy:
Last time I checked (2 years ago), VideoEncoder did not support transparent video for any codec. Adding it to WebM is easy but somewhat pointless if it can't be encoded.

I'll need to check again if it's possible now in 2025! If it is, I'll be sure to add it to Mediabunny.

Vanilagy:
<img width="453" height="57" alt="Image" src="https://github.com/user-attachments/assets/9e66a5be-e4b0-490e-ab5d-1291d408d4c7" />

Okay no, still not possible in 2025. I tested it with this:
```ts
const canvas = document.createElement('canvas');
const context = canvas.getContext('2d');

const frame = new VideoFrame(canvas, {timestamp: 0});

const encoder = new VideoEncoder({
  output: console.log,
  error: console.error,
});

console.log("soda pop")

encoder.configure({
  codec: 'vp09.00.10.08',
  width: canvas.width,
  height: canvas.height,
  bitrate: 1e6,
  alpha: 'keep'
})
encoder.encode(frame);
```

Makes sense that it fails. Here in the Chromium source code:
<img width="575" height="388" alt="Image" src="https://github.com/user-attachments/assets/a0657c2d-9306-4907-b2e2-0e139e969bf8" />

In Firefox, the encoder goes through but decoding the encoded chunk reveals it has no alpha:
<img width="460" height="239" alt="Image" src="https://github.com/user-attachments/assets/616467ab-a296-4720-8689-92b4c0e1daf9" />

Until VideoEncoder doesn't support transparency, it doesn't feel worth it to me to add this to Mediabunny.

shealan:
That's a shame. Thanks for looking into it. Hopefully it will be supported eventually.

pucelle:
Meets same problem, thanks.

Vanilagy:
I actually stand corrected! Turns out, WebM transparency works different from how I thought it worked. It is actually packed into a separate codec stream alongside the normal color data:

[...]

PR #94, opened 2025-09-04
Experimental WebM alpha support

cr3314:
Should suffice for: https://github.com/Vanilagy/mediabunny/issues/55.
WebM alpha works by adding additional data, which is another encoded stream. [Implementation ref](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/libvpxdec.c)

Only support for VP8/VP9 for encoding.
VP8/VP9/AV1 for decoding.
AV1 with alpha in WebM seems uncommon that no browser support playback with alpha natively. However, the basically same logic for AV1 case, [alpha support with AVIF is common](https://aomediacodec.github.io/av1-avif/#av1-alpha-image-item).

Note: Chrome support WebM with VP8/VP9 alpha encoding natively with MediaRecorder. [WebRTC Sample](https://webrtc.github.io/samples/src/content/capture/canvas-record/)

Changes:
- [x] 1. Minor additions to matroska-demuxer and matroska-muxer
- [x] 2. Custom alpha decoder + combining.
    - [x] Demuxer detect feature need, (probably better to do at `CustomVideoEncoder.supports`, but no access to packet now)
    - [ ] (*) Scope limitation: software decoding only
- [x] 3. Custom alpha encoder + extracting.
    - [x] Fix alpha has limited range
    - [ ] (*) Main library refactor needed: detect feature need based on output format instead of always override default behaviour when custom encoder registered.
- [x] 4. Handle YUV planar format.
- [x] 5. Handle RGB format, from `canvas` / Firefox's `VideoDecoder` result.
    - [ ] (*) Limitation: per pixel manipulation, inefficient compare to YUV planar.
    - [ ] (*) Limitation on transcoding: Chrome's `VideoDecoder` gives `VideoFrame.format: null` for HEVC with alpha. Might be possible to handle by `copyTo` BGRX formats. But [`VideoFrameCopyToOptions`](https://www.w3.org/TR/webcodecs/#videoframe-copyto-options) is not supported in `VideoSample` now. But can hope that Chrome fixes it on their side. [Example HEVC with alpha file](https://jakearchibald.com/c/hevc-D_-LfvxD.mov).
- [x] 6. Add alpha handling for internal canvas usage.
- [x] 7. Check and resolve irregularity for Chrome, Firefox, Safari.
- [x] 8. Check if encoded webm works in `MediaSource`: [Chromium needs `BlockAddID` to come first](https://source.chromium.org/chromium/chromium/src/+/main:media/formats/webm/webm_cluster_parser.cc;l=404;drc=48d6f7175422b2c969c14258f9f8d5b196c28d18;bpv=1;bpt=1) when used in MSE.

Technically, only 1. is needed in the library for user to implement their custom support (without forking).
I see that a "basic" support could be useful for many so I created the PR. But it is totally fine to close this if there are other plans (or if the code is an abomination).

IMPORTANT NOTE on 3.: `VideoFrame` with RGB formats like `canvas` would get clamped to `16-255` range internally in browser `VideoEncoder` probably due to conversion to `'I420'` but a full range `'J420'` is needed for alpha channel. While browser does not break manually created `'I420'` `Videoframe` now, this whole manual separate alpha in JS implementation idea relies on this piece of browser implementation detail to work.

Note:
HEVC with alpha decoding already works as-is, but internal `canvas` usage presumes always opaque (fixed with 6.).

Not in scope:
(*) Items are out of scope due to main library impacts/significant extra complexity while not required for the feature.
- [x] 1. Investigate on Firefox slow VP9 encoding (generally). https://github.com/Vanilagy/mediabunny/issues/102#issuecomment-3265573488
- [ ] 2. Encode AV1 with alpha. Descoped, no native decoding support in any browser.
- [ ] 3. Optimization: `copyTo` from hardware backed `VideoFrame` by [async readback](https://issuetracker.google.com/issues/340606792). e.g. `encode` for `CanvasSource` can be faster by parallel readback instead of simply `await extractAlpha`.
- [ ] 4. Should we split the `bitrate` to maybe main 3 : 1 alpha ratio? Seems not performed in `FFmpeg` but quite wasteful without for `bitrateMode: 'constant'`.

### DIFF
--- a/dev/convert.html
+++ b/dev/convert.html
@@ -5,6 +5,8 @@
 
 <script type="module">
 	MediabunnyMp3Encoder.registerMp3Encoder();
+	Mediabunny.registerWebMSeparateAlphaDecoder();
+	Mediabunny.registerWebMSeparateAlphaEncoder();
 
 	const fileInput = document.createElement('input');
 	fileInput.type = 'file';
@@ -97,7 +99,8 @@
 			*/
 			/*
 			video: {
-				codec: 'avc',
+				codec: 'vp9',
+				alpha: 'keep',
 			},
 			audio: {
 				codec: 'mp3',
--- a/docs/guide/media-sources.md
+++ b/docs/guide/media-sources.md
@@ -74,7 +74,8 @@ type VideoEncodingConfig = {
 - `hardwareAcceleration`: A hint that configures the hardware acceleration method of this codec. This is best left on `'no-preference'`.
 - `scalabilityMode`: An encoding scalability mode identifier as defined by [WebRTC-SVC](https://w3c.github.io/webrtc-svc/#scalabilitymodes*).
 - `contentHint`: An encoding video content hint as defined by [mst-content-hint](https://w3c.github.io/mst-content-hint/#video-content-hints).
-- `sizeChangeBehavior`: Video frames may change size overtime. This field controls the behavior in case this happens. Defaults to `'deny'`. 
+- `sizeChangeBehavior`: Video frames may change size overtime. This field controls the behavior in case this happens. Defaults to `'deny'`.
+- `alpha`: Alpha option defined by [alpha-option](https://w3c.github.io/webcodecs/#alpha-option). Defaults to `'discard'`.
 - `onEncodedPacket`: Called for each successfully encoded packet. Useful for determining encoding progress.
 - `onEncoderConfig`: Called when the internal encoder config, as used by the WebCodecs API, is created. You can use this to introspect the full codec string.
 
--- a/docs/guide/packets-and-samples.md
+++ b/docs/guide/packets-and-samples.md
@@ -91,6 +91,7 @@ constructor(
     duration: number, // in seconds
     sequenceNumber?: number,
     byteLength?: number,
+    additions?: Uint8Array, // Additional data, currently only for Matroska format
 );
 ```
 
--- a/docs/guide/supported-formats-and-codecs.md
+++ b/docs/guide/supported-formats-and-codecs.md
@@ -61,7 +61,7 @@ Mediabunny ships with built-in decoders and encoders for all audio PCM codecs, m
 
 Not all codecs can be used with all containers. The following table specifies the supported codec-container combinations:
 
-|                |   .mp4   | .mov  | .mkv  | .webm[^1] | .ogg  | .mp3  | .wav  | .aac  |
+|                |   .mp4   | .mov  | .mkv  | .webm[^1][^2] | .ogg  | .mp3  | .wav  | .aac  |
 |:--------------:|:--------:|:-----:|:-----:|:---------:|:-----:|:-----:|:-----:|:-----:|
 | `'avc'`        |    ✓     |   ✓   |   ✓   |           |       |       |       |       |
 | `'hevc'`       |    ✓     |   ✓   |   ✓   |           |       |       |       |       |
@@ -87,11 +87,12 @@ Not all codecs can be used with all containers. The following table specifies th
 | `'pcm-f64be'`  |    ✓     |   ✓   |       |           |       |       |       |       |
 | `'ulaw'`       |          |   ✓   |       |           |       |       |   ✓   |       |
 | `'alaw'`       |          |   ✓   |       |           |       |       |   ✓   |       |
-| `'webvtt'`[^2] |   (✓)    |       |  (✓)  |    (✓)    |       |       |       |       |
+| `'webvtt'`[^3] |   (✓)    |       |  (✓)  |    (✓)    |       |       |       |       |
 
 
 [^1]: WebM only supports a small subset of the codecs supported by Matroska. However, this library can technically read all codecs from a WebM that are supported by Matroska.
-[^2]: WebVTT can only be written, not read.
+[^2]: WebM alpha channel support [can be enabled (experimentally)](#webm-with-alpha-channel-experimental)
+[^3]: WebVTT can only be written, not read.
 
 ## Querying codec encodability
 
@@ -321,3 +322,43 @@ All instance methods of the class can return promises. In this case, the library
 ::: warning
 The samples passed to `onSample` **must** be sorted by increasing timestamp. This especially means if the decoder is decoding a video stream that makes use of [B-frames](./media-sources.md#b-frames), the decoder **must** internally hold on to these frames so it can emit them sorted by presentation timestamp. This strict sorting requirement is reset each time `flush` is called.
 :::
+
+## WebM with alpha channel **(Experimental)**
+
+While WebM video format and its video codecs support [alpha channel](https://wiki.webmproject.org/alpha-channel), there is [no native WebCodecs support](https://github.com/w3c/webcodecs/issues/200) as of writing.
+A set of custom coders is provided to support it by separately handle the alpha channel of the input.
+
+::: warning
+It is experimental and still requires more work on the main library and CustomCoder APIs.
+Issues examples:
+- No proper way for user to confirm if a file is indeed a WebM with alpha and thus if opt-in support is needed.
+- CustomCoder supports detection is not specific enough with current abstraction.
+:::
+
+### Encoding WebM with alpha channel
+
+Support can be added with `registerWebMSeparateAlphaEncoder`.
+Then, opt in with `videoEncodingConfig.alpha = 'keep'` on AV1/VP9/VP8 codecs.
+
+::: info
+While AV1 with alpha muxing works, browsers wouldn't decode natively in browser.
+:::
+::: warning
+The alpha channel support is part of WebM file format and useful to encode in the **WebM/MKV muxer only**.
+While it should not cause problem to opt-in for 'keep' in other formats, the alpha would still be discarded in other formats.
+:::
+
+### Decoding WebM with alpha channel
+
+Support can be added with `registerWebMSeparateAlphaDecoder`.
+After registeration, Matroska demuxer would then detect if actual usage is needed for convenience. Make sure this is registered before creating `Input` of `WebM`!
+It would only output a frame with alpha if a particular packet to contains the additional data. Packets from the internal WebM/MKV demuxer would include it.
+```ts
+const packet = new EncodedPacket(
+	...,
+	// Uint8Array, the alpha encoded data
+	// Will be included automatically if using file-based `Source` + `VideoSampleSink`
+	// Can also be added when manually creating EncodedPacket
+	additions,
+)
+```
--- a/docs/guide/supported-formats-and-codecs.md
+++ b/docs/guide/supported-formats-and-codecs.md
@@ -339,15 +339,14 @@ Issues examples:
 ### Encoding WebM with alpha channel
 
 Support can be added with `registerWebMSeparateAlphaEncoder`.
-Then, opt in with `videoEncodingConfig.alpha = 'keep'` on AV1/VP9/VP8 codecs.
+Then, opt in with `videoEncodingConfig.alpha = 'keep'` on VP9/VP8 codecs.
 
 ::: info
-While AV1 with alpha muxing works, browsers wouldn't decode alpha natively in browser.
-Also, Safari does not support WebM alpha native.
+Safari does not support WebM alpha natively.
 :::
 ::: warning
 The alpha channel support is part of WebM file format and useful to do alpha encoding process in the **WebM/MKV format only**.
-While it should not cause problem to opt-in for 'keep' in other formats, the alpha would be wasteful in other formats.
+While it should not cause problem to opt-in for 'keep' in other formats, the alpha processing would be wasteful in other formats.
 :::
 
 ### Decoding WebM with alpha channel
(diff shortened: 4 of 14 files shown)


Vanilagy:
Interesting that you're tackling, I'll take a look at this another time.

ghost:
Take your time. Thanks for your time!
Personally don't want to change/refactor a lot of the main library so there are some limitations commented in code.

# Some rough testing:
[Test file, 720p 60fps 9s](https://rotato.netlify.app/alpha-demo/movie-webm.webm)
Using the dev/convert.html
Result looks a bit different. Chrome gives smpte170m, Firefox gives bt709. Chrome looks closer to original.
| Settings | Description |
| --- | --- |
| `{ video: { codec: 'vp9', forceTranscode: true } }` | baseline |
|`registerWebMSeparteAlphaDecoder()` | ^add alpha decoding |
|`{ video: { ..., alpha: 'keep' } }` | ^add alpha encoding |
|`{ video: { ..., rotate: 90, fit: 'contain' }}` | ^add rotate to force re-render |
|`{ video: { ..., alpha: 'discard' }}` | ^remove encode alpha |

Different resolution test, upscaled by FFmpeg:
| Original HD | FHD | UHD |
| --- | --- | --- |

## Macbook m1 air

### Chrome v140:
| HD | FHD | UHD |
| --- | --- | --- |
| ~1.4s | ~2.3s | ~7.2s |
| ~1.4s | ~2.5s | ~9.5s |
| ~1.7s | ~3.5s | ~13.0s |
|  ~3.8s | ~6.6s | ~24.3s |
| ~1.8s | ~2.9s | ~11.0s |

### Firefox v142:
| HD | FHD | UHD |
| --- | --- | --- |
| ~17.8s | ~33.4s | ~64.9s |
| ~18.1s | ~34.2s | ~74.8s |
| ~19.1s | ~37.9s | ~103.3s |
| ~25.9s | ~37.6s | ~125.1s |
| ~24.9s | ~30.7s | ~87.4s |

Note: native `VideoEncoderConfig.alpha: 'keep'` reports supported, but it is actually [simply ignored](https://bugzilla.mozilla.org/show_bug.cgi?id=1909379).

### Safari v18.6
| HD | FHD | UHD |
| --- | --- | --- |
| ~1.5s | ~2.2s | ~7.7s |
| ~1.2s | ~1.4s | ~4.9s |
| ~2.4s | ~4.1s | ~14.8s |
| ~6.1s | ~11.5s | ~39.4s |
| ~3.6s | ~6.1s | ~19.0s |

Note: setting 2 has BROKEN OUTPUT, `VideoFrame.format: 'I420A'` not handled correctly in native VP9 `VideoEncoder`.

Edit: better format

ghost:
Fixed for Safari, fluffs:
Apparently Safari, when `VideoFrame.duration` is provided, it wouldn't output original timestamp.
Somehow Safari `context.globalCompositeOperation = 'copy'` doesn't work correctly. Add back the `clearRect` logic, (can also consider `context.reset` to also reset transformation).

Interested on why is `HTMLCanvasElement` preferred over `OffscreenCanvas` in the library? I think `OffscreenCanvas` should have better performance in general by not required tie to UI thread.

Vanilagy:
Will take a closer look at this PR next week. Done enough Mediabunny this week.

Vanilagy:
Just for the record, if this is still a work-in-progress, please label it as a draft PR!

ghost:
Sorry for confusion. It is ready/usable since I opened the PR with changes up to [2667d38](https://github.com/Vanilagy/mediabunny/pull/94/commits/2667d38d48f58af2538b12cd4fb4aa08fc478d2d). But sadly, I cannot create a stacked PR from a fork.
Also, I have rebased everything due to `index.ts` merge conflict. This is my mistake. Definitely should have used merge to keep the history.

The changes after opening the PR are Safari specific fixes.
Originally, I simply disable support Safari due to unexpected problems with encoding and Safari not supporting decoding of WebM with alpha natively. But end up able to fix it so I included the fix and some minor improvements.

The remaining items in list are not necessary for the feature and are for future reference.

ghost:
@Vanilagy Re-request review is definitely wrong. Would like to say sorry that I usually think about and problem repeatedly with scattered schedule and immediate testing and end up editing the comment too many times. Now, I think it is reasonably complete for you to read.
While I am also happy to help on implementation, I guess it would be easier for your to take over. Please let me know your preference.
Will definitely try to review if you want to continue on this and make some new changes.
This PR has already been quite messy. It is also fine to close this PR and start over.
Definitely glad if this could help you! Let me know if you need more of my help.

I don't really know much more than how it basically works by having 2 streams before starting honestly. Implementing the logic in a basic way is simple but there is quite a lot of weirdness in browsers that I end up dealing with. Also partly the reason why I have to keep editing.

Thanks for your detailed comments. I definitely agree with your ideas.
Well, some are the results of my personal preference to avoid making intrusive changes to core parts of the main library and changing default behavior thus I chose to implement it as a `CustomEncoder`. So that the main library usage would not easily get regression issue without explicit opt-in. Also, I believe it is easier to integrate later than separate. The design is definitely poor partly due to not integrating.

Relies in same order:
- Sure. `VideoEncodingConfig.alpha?: 'keep' | 'discard' = 'discard'` is simply mirroring WebCodecs spec [VideoEncoderConfig.alpha](https://w3c.github.io/webcodecs/#enumdef-alphaoption).
- Sure. The naming of `EncodedPacket.additions` field: I simply followed the original `additions` field from WebM subtitle implementation in mediabunny.
- 
    - I agree that it is fine to check if the first sample successfully give alpha output in a real use case.
    - On decoding behavior, I think better follow [a browser implementation](https://source.chromium.org/chromium/chromium/src/+/main:media/filters/vpx_video_decoder.cc;l=348;drc=48d6f7175422b2c969c14258f9f8d5b196c28d18?q=vpxVideodecoder&ss=chromium%2Fchromium%2Fsrc). So that there is fewer potential questions like "my file does not work the same" to mediabunny side.
- [`EncodedVideoChunkMetadata.alphaSideData` is in spec](https://www.w3.org/TR/webcodecs/#encoded-video-chunk-metadata) but there is no real implementation yet. I personally agree that the alpha data should be on the packet. It is a bit awkward on the `metadata` since decoding side spec don't have a similar thing.
- 
    - Also, relates to the spec, since encoding side is in spec. We can expect browser to [eventually provide native support](https://bugzilla.mozilla.org/show_bug.cgi?id=1909379) and maybe we should detect support for it. Note that Firefox does not report `supported` correctly.
    - But on decoding side, spec seems lacking and thus a custom implementation is probably still needed. Originally I thought there are chances that both will be supported natively at the end.
- I am not too sure how you would implement it to the main `VideoEncoderWrapper`. I personally still tends to create a e.g. `AuxiliaryVideoEncoderWrapper` and wrap 2 native `VideoEncoder` [(and 3 for AVIF alpha + depth)](https://aomediacodec.github.io/av1-avif/#av1-alpha-image-item). `VideoEncoderWrapper` detect usage and create `AuxiliaryVideoEncoderWrapper` or native `VideoEncoder`. `AuxiliaryVideoEncoderWrapper` could implement `interface IVideoEncoder extends VideoEncoder {}`. Since the output synchronization logic is not needed for other cases, I think it is better to make it stay independent to avoid overhead for other cases.
- I had experimented with hardware acceleration flow earlier but I would think it would make most sense to do so only in `CanvasSink` because user should know about it when additional GPU processing is added.
    - Extract alpha: actually most `VideoEncoder` still read back the GPU backed `VideoFrame` internally. [e.g. chromium](https://issues.chromium.org/issues/40162806). Also, the situation in the "IMPORTANT NOTE on 3" above, it is necessary to manually create a `'I420'` `VideoFrame` to make alpha encoding work properly in full 0 - 255 range, but in these GPU APIs, there is [no way to create YUV texture](https://www.w3.org/TR/webgpu/#enumdef-gputextureformat). Meaning that, read back and use JS is indeed necessary now. [JS fiddle, comment inside for further explaination](https://jsfiddle.net/kw4b5ud8/). Try on chrome.
    - Combine alpha: it is possible with canvas2d (only on UI thread). [See JS fiddle above](https://jsfiddle.net/kw4b5ud8/). Performance was indeed much better on Firefox on a >4k video but it is still not reaching 60fps for my Macbook M1 air. Chromium's YUV based simple `uint8.set` is already performant enough imo. But afaik about canvas2d, there is no other way to convert colour to alpha channel. So, WebGL / WebGPU implementation is needed for hardware acceleration. However, it will add quite a bit of boilerplate to the library.
        - Realistically, if I were to develop a graphical application which is better in a pure hardware accelerated flow, I would rather have the main and alpha frames directly given to me and I perform my own composition to avoid intermediate canvas processing overhead. So, maybe, a `VideoDecodingConfig.alpha: 'separate'` option could also be useful.
- The native implementations I have seen are fully synchronous and CPU only and there is no good reference implementation on the dropout problem.
    - Fully synchronous that means it is keeping the frame and then when after main encoder has output, add and wait for alpha encoder (e.g. by with a separate `latencyMode: 'quality'`) to ensure 100% match. But this is probably very slow. The latency would be even higher than just use `'quality'`.
    - Well, simple way is to just only support `'quality'` to ensure no frame drop. I don't think anyone would realistically want an output that could randomly go with or without alpha back and forth. Could opt for `'vp8'` for lower latency.
    - When there is a frame drop of the main encoder but alpha encoder didn't drop the same frame, the packet sequence for alpha will be broken. Decoding alpha will then fail. Need to manually omit all the alpha data to fallback to opaque and avoid decoder failure until next keyframe is received to make the file "fine" to play through in browsers. In `VideoEncoder.output` callbacks, could iterate the `Map` and handle the skipped alpha packets by current `EncodedVideoChunk.timestamp` check against `mapKey`. Discard previous alpha data if main has dropped frame. `resolve(null)` if alpha has dropped frame.  See new testing code below.
    - ^ Note: Safari might not return you the same `microsecondTimestamp` that is passed from `VideoFrame.timestamp` when a `VideoFrame.duration` is provided so it is intentionally left out in my code. Forgot to comment it in the code to ensure mapping.
    - ^^ Note: the current logic to immediately `resolve(null)` on frame without alpha to extract is wrong! If a user really just pass a frame without alpha in the middle of alpha encoding. Should encode a full opaque alpha frame for completeness.

ghost:
^ Here is some testing code on what last point and dropout would look like: https://github.com/cr3314/mediabunny/compare/webm-alpha...cr3314:mediabunny:webm-alpha-test-dropout
In short, if alpha output does not drop the exact same frames. It is not realistic to use.

Vanilagy:
Thanks for the reply, don't worry about editing, it's a complex feature.

I would say it's best to move forward with a new PR which aims to implement this feature in a non-experimental way. I still think it's important for Mediabunny to be able to do this, and I would love if you're the main author of this new PR! I think you understand the topic very well and we're aligned on the vision for the API, plus I'm working on other features so I appreciate any help I can get.

---

Regarding the implementation in `VideoEncoderWrapper`, I'm not sure how hard it would be exactly. The alpha logic would basically need to kick whenever the `VideoEncoder` branch is hit (and `encoder` is set). Perhaps just adding a new `alphaEncoder` field will be enough, and then based on if that's set or not (which is determined at init time), the encode logic changes behavior. I would of course extract the alpha-related stuff into separate methods though, so that the normal code path isn't cluttered. I think it's best to just start and see if any rough edges are encountered. I'm fine with auxilliary classes that help with the synchronization.

Regarding hardware acceleration, I don't mind using GPU acceleration even in the non-canvas case. The frames are in VRAM most of the time anyways, and Mediabunny's focus is on client-side execution where these primitives are available anyway (even in Workers!) I do agree though, if there is a way to do it super efficiently in CPU (using a single Uint8Array.set), then maybe that's beneficial. Although it's also cool to have a single code path that behaves the same for all cases, less mental overhead that way.

I wouldn't mind a bit of WebGL code, just for a simple alpha combine, I don't think this will need substantially more than 50 LOC. Should be very simple, just a single full-screen rect being drawn with a trivial vertex shader and then a slightly more complex fragment shader. Would not use WebGPU, overkill for this.

We could consider `VideoDecodingConfig.alpha: 'separate'`. There I guess there should be an `alphaSample` property on `VideoSample`? That's how I could imagine the API to be like. Although by default, again, I think the library should "just" handle alpha and make sure its baked into the VideoSample automatically, since this meshes nicely with any other code that consumes samples from Mediabunny.

Oh and of course, like you've already done, we need to make sure that any derived APIs like CanvasSink also are extended with alpha functionality.

Regarding the dropout, maybe it's fine to just force latencyMode: 'quality' on the alpha encoder then? Not sure. Best not to make it too complex if the use case is niche.

---

So, feel free to open a new draft PR and start hacking on this, I see this as a very important feature for Mediabunny and I really appreciate you pushing forward to make this a reality.

One thing I might want to do before: I want to to extract `VideoEncoderWrapper` and `VideoDecoderWrapper` (and the audio variants too) into publicly exposed API that you can use standalone. This should be mostly copy-paste but will require a few adjustments at least. And then, these would live in separate files and won't be together with the media sources and media sinks. I will be doing this in the next couple of days, so it might be smart to wait before starting your new PR to avoid conflicts. But you can also just start and then accept a few merge conflicts later, whatever you prefer.+

Closing to make room for a new PR.

PR #97, opened 2025-09-05
Add Crop feature

pablocubico:
This PR is quite small (apart from the crop example files).

It affects `conversion.ts` and `media-sink.ts`:

* adds & validates the new cropping options (top, left, width, height)
* calculates the final width/height (depending on whether "crop" was set or not)
* renders the frame into an HTML Canvas or OffscreenCanvas.
* defaults to black (`#000`) for pixels beyond frame (maybe that could be an option too?)

✅ tests passing
✅ build working (had to add a few docblocks)

Hope this feature helps (it will definitely help me!)

### DIFF
--- a/src/conversion.ts
+++ b/src/conversion.ts
@@ -228,7 +228,7 @@ const validateVideoOptions = (videoOptions: ConversionVideoOptions | undefined) 
 		throw new TypeError('options.video.rotate, when provided, must be 0, 90, 180 or 270.');
 	}
 	if (videoOptions?.crop !== undefined) {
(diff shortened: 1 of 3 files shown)


Vanilagy:
Thank you for this! I will look at this PR next week.

luisrudge:
according to claude, this is the only missing feature for my project to move from ffmpeg to mediabunny. Following this closely!

luisrudge:
@Vanilagy thanks for working on this! examples are useful because llms use them to modify the function call to custom scenarios

PR #101, opened 2025-09-07
Add dynamic video URL support for Media Player

jebin2:
This update allows the MediaBunny player to load videos dynamically via a URL query parameter. Users can now pass a `video_url` in the page URL to load any video without modifying the code.

Makes it easier for demo video.

**Usage Example:**

```
https://mediabunny.dev/examples/media-player/?video_url=https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4
```

### DIFF
--- a/examples/media-player/media-player.ts
+++ b/examples/media-player/media-player.ts
@@ -694,3 +694,12 @@ document.addEventListener('drop', (event) => {
 		void initMediaPlayer(file);
 	}
 });
+
+document.addEventListener("DOMContentLoaded", () => {
+	const urlParams = new URLSearchParams(window.location.search);
+	const url = urlParams.get("video_url");
+
+	if (url) {
+		void initMediaPlayer(url);
+	}
+});


Vanilagy:
Can you elaborate more on the intended use case for this? Since this is meant as an example, not a sharable video player.

Not too against adding this, but if we do, then at least it should be consistent and added to all examples that require an input file.

jebin2:
My main use case is wanting an easy way to share playable demo videos for projects (on GitHub, Hugging Face, or just via direct links) without editing the code each time. Passing a `video_url` query param made it much simpler to load a video and share a demo link.

And I completely understand that this project is meant as an example rather than a sharable player. If it’s not aligned with the project’s purpose, no problem at all, i totally understand and I’m happy to close this PR.

Alternatively, if you’d like, I can adjust the approach to apply consistently across examples that take input files.

I really appreciate the review!

Vanilagy:
I understand! I will close this PR for now since this feature doesn't really align with these being examples and not actual apps. But, thanks for the feedback, and maybe there will be an actual Mediabunny player website one day.

PR #109, opened 2025-09-13
Multiple inputs support for Conversion

bardcrab:
I propose to add functionality to handle multiple inputs in Conversion. This feature may be useful when there are need to mux audio and video streams which are stored each in separate files. I'm trying to make use of mediabunny in FreeTube project to download and mux audio and video streams to single file. I've tested tweaked version of mediabunny with my changes and it works very well.

There are another unrelated change in this pull request: I noticed that when used with Electron the mediabunny is unable to load because of unreachable node fs module, requested in source.ts to query file size. I added a fix for this, a simple stub that checks if it is running under Electron and then avoids access to fs. This fix is far from ideal, but without it the whole library refuses to load.

### DIFF
--- a/docs/guide/converting-media-files.md
+++ b/docs/guide/converting-media-files.md
@@ -45,7 +45,25 @@ await conversion.execute();
 // output.target.buffer contains the final file
 ```
 
-That's it! A `Conversion` simply takes an instance of `Input` and `Output`, then reads the data from the input and writes it to the output. If you're unfamiliar with [`Input`](./reading-media-files) and [`Output`](./writing-media-files), check out their respective guides.
+Additionally, it is possible to pass multiple `Input`'s in `input` field as array:
+```
+// ...
+
+const inputA = new Input({ ... });
+const inputB = new Input({ ... });
+const output = new Output({
+	format: new WebMOutputFormat(),
+	target: new BufferTarget(),
+});
+// This may be useful in case when audio and video streams are stored in separated files
+
+const conversion = await Conversion.init({ input: [ inputA, inputB ], output: output });
+await conversion.execute();
+```
+
+When multiple inputs are used they duration must be equal. Also, if inputs contain metadata, only the metadata from the first provided input will be used in the output.
+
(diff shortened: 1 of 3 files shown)


Vanilagy:
Thanks for the PR! I'll be leaving this PR open for now because I want to make some adjustments to this feature, but I just don't want to do it right now. This feature has been on my roadmap for a while, and it's important, but I want to make sure it's done right. I'm sure it works well for your current use case, but I also want to enable "concat" operations that equally require multiple inputs as well. So, you could basically use multiple inputs to either merge them together or to concatenate multiple files together - and there needs to be API to distinguish the two. Hence, I don't wanna merge this before I have that API figured out to avoid future breaking changes.

Regarding the Electron thing: Can you elaborate? I wrote it to be a dynamic import so that if the code isn't run, nothing is imported. So why is Electron still trying to resolve it?

bardcrab:
Got it, thanks for feedback.

Regarding the Electron and dynamic import: I see that this code does not get actually executed when application start. But looks like there are some precompilation/packaging step which checks all the code for unconditional dynamic imports. Here is the calls trace:

```
WARNING in ./node_modules/mediabunny/dist/modules/src/source.js 370:33-74
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/mediabunny/dist/modules/src/index.js 17:0-127 17:0-127 17:0-127 17:0-127 17:0-127 17:0-127 17:0-127 17:0-127
 @ ./src/renderer/store/modules/downloads.js 3:0-111
 @ ./src/renderer/store/index.js 6:0-44 19:4-13
 @ ./src/renderer/main.js 5:0-34 40:2-7
```

The error is marked as a warning, but it block application completely.

Vanilagy:
Oh interesting. So it fails because the import is not known at compile time. But still strange that it checks it at all, given that it's a dynamic import. Annoying??

Vanilagy:
This warning should disappear in the next version, I've made some changes

Vanilagy:
Oh the "is electron" thing isn't needed anymore btw, that's fixed

absidue:
> Oh interesting. So it fails because the import is not known at compile time. But still strange that it checks it at all, given that it's a dynamic import. Annoying??

Bundlers trace all imports because they need to know what files to include in the output, if they skipped all dynamic imports `import(./some-file.js')` wouldn't work at runtime because the file wouldn't be copied into the output.

As for `process.env.IS_ELECTRON`, that is a FreeTube specific build flag, so it never should have been included here.

PR #122, opened 2025-09-17
WebM alpha support

ghost:
For: https://github.com/Vanilagy/mediabunny/issues/55. Basically the same feature and performance as https://github.com/Vanilagy/mediabunny/pull/94.

API: basically follow WebCodecs spec.
- Outputting to/using `meta.alphaSideData` as if browsers support it.
- Since the aim is to supply the `Wrapper` publicly, the logic is rewritten to implement the same `interface` for `VideoEncoder` and `VideoDecoder`.
- Alpha wrapper to consume the output from native coders as output iterators in a while-loop. At least the logic could be read linearly.

Hardware acceleration: still no, I will leave the fun for others.

Hope that there are no major issues with the implementation.

PS:
I will be on holiday for a while. So, if there is no major issues, would definitely appreciate help on minor edits. ❤️

### DIFF
--- a/src/matroska/matroska-demuxer.ts
+++ b/src/matroska/matroska-demuxer.ts
@@ -142,7 +142,7 @@ type ClusterBlock = {
 	data: Uint8Array;
 	lacing: BlockLacing;
 	decoded: boolean;
-	additions?: Uint8Array;
+	additions: Uint8Array | null;
 };
 
 type CuePoint = {
@@ -844,8 +844,8 @@ export class MatroskaDemuxer extends Demuxer {
 					referencedTimestamps: originalBlock.referencedTimestamps,
 					data: frameData,
 					lacing: BlockLacing.None,
-					additions: originalBlock.additions,
 					decoded: true,
+					additions: originalBlock.additions,
 				});
 			}
 
@@ -1373,6 +1373,7 @@ export class MatroskaDemuxer extends Demuxer {
 					data: blockData,
 					lacing,
 					decoded: !hasDecodingInstructions,
+					additions: null,
 				});
 			}; break;
 
@@ -1415,6 +1416,7 @@ export class MatroskaDemuxer extends Demuxer {
 					data: blockData,
 					lacing,
 					decoded: !hasDecodingInstructions,
+					additions: null,
 				};
 				trackData.blocks.push(this.currentBlock);
 			}; break;
@@ -1440,8 +1442,6 @@ export class MatroskaDemuxer extends Demuxer {
 			case EBMLId.BlockAddID: {
 				if (readSignedInt(slice, size) === 1) {
 					this.readContiguousElements(slice.slice(dataStartPos, size));
-				} else if (this.currentBlock?.additions) {
-					this.currentBlock.additions = undefined;
 				}
 			}; break;
 


Vanilagy:
You're an absolute champ

Vanilagy:
You already on holiday? 😂

ghost:
Yes. and I am not going to try coding on phone anymore. 😸 
Just suddenly realize that `EventTarget` exists before bed.

Vanilagy:
Take a laptop bro!! Where you doing vacation?

Vanilagy:
I ended up reimplementing this in #145. I used this PR as reference a lot, especially the insights about YUV and needing to create the frame in software so that the color spaces don't end up all fucked up.

I felt some of the solutions in this PR were a bit too hacky/clever and I wanted something simpler. Especially with regards to the encoder interfaces, the mocks and the promise-heavy logic to sync up decoders/encoders. I ended up solving the syncing problem with a queue-based approach instead. The merging/splitting I did purely hardware-accelerated using WebGL 2, with some crazy YUV hacking for the alpha encoding actually.

Hope you're not mad about this! This PR, the previous PR and all the discussion around this topic were instrumental for bringing this feature to life.

If you got time, maybe you could check out my implementation and bombard it with test files of yours. It appears quite robust to me, doing well with videos that switch between alpha and no alpha, and also with reading videos where the alpha side data has a different key/delta rhythm than the main data. (This is something you didn't address in your logic afaik). I made sure when encoding that the key/delta rhythms are synced up. To do this, I had to wire the two encoders in series instead of in parallel: Only when the color part is encoded do I start encoding the alpha, because I need to force a key frame if the color came out as a keyframe.

ghost:
@Vanilagy
Definitely no problem! Great work! My original goal was to just build a reference implementation anyway.

Yea, the sync by iterators is quite ridiculous. But it is the most (linearly) readable way for me on how would the parallel sync logic could actually work and reasonably support all those edge cases.
But I do think it is better to implement the `VideoEncoder` and `VideoDecoder` interface because that allows the alpha wrappers to be used as if they are the native `VideoEncoder` and `VideoDecoder`. Ofc, there is no native way for `VideoDecoder` so I had to add an optional parameter to the API.
- While it is definitely a non concern for Mediabunny, it allows for separately publishing the alpha wrappers standalone for anyone to integrate (potentially) seamlessly.
- My belief is that browsers would eventually support alpha natively. So, it is better to avoid breaching into the core code for easily opt-out or even removal in the future.

> alpha side data has a different key/delta rhythm than the main data. (This is something you didn't address in your logic afaik).

From my understanding, the alpha and main must have keyframes at the same block otherwise seeking will result in decoder failure, at least in all native implementation. So, I only handled the case when alpha delta have gaps. I just added an arbitrary limitation to my code because it is how browser works and it sounds quite expected to me.

>  I did purely hardware-accelerated using WebGL 2, with some crazy YUV hacking for the alpha encoding actually.

I avoided hardware acceleration myself as I thought that the rabbit hole could be fairly deep but I really should have noted these down when imagining.
1. Be careful that browsers have [a hard limit (as low as 8 across all `window`) on the total WebGL context counts](https://stackoverflow.com/questions/61277222/how-to-determine-number-of-active-webgl-contexts). I personally would try to use a WebGL singleton to reduce chance of interfering with the users' WebGL application when multiple `VideoDecoderWrapper` / `VideoEncoderWrapper` are created. The users will certainly come to you when they find their WebGL application running into context lost.
2. You should probably [`gl.readPixels` in a non blocking way](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_non-blocking_async_data_readback) or [check three.js implementation](https://github.com/mrdoob/three.js/pull/28291). It could be blocking for longer than manually assigning Y plane from RGBA in JS, interfering with the main application and nullifying the gain from using hardware acceleration, [demo](https://kai.graphics/getBufferSubDataAsync-Demo/). But yea, `async` would make things much more complicated and you might eventually found yourself in the same boat with my implementation lol.
3. You could probably omit the alpha shader by `texImage2D` with format `gl.ALPHA` and get the full Y plane already and then `uint8.fill` the UV after `readPixels` (similarly for merging, could use the original `gl.LUMINANCE` value without converting to RGB by browser internally).
    - The recreating the `VideoFrame` with opaque format for "color" is only needed for a browser bug in Safari for `'I420A'`, might also be already fixed in latest version.
    - Instead of re-rendering in WebGL, You could probably simply pass the original `VideoFrame` (or recreate by `new VideoFrame` with opaque format) for encoding color since you have merge into a RGBA. Unlike my combined `I420A` case. But, you might get result `chunk.timestamp` off-by-1 issue on Safari if not recreated. Hopefully they might have fixed it on `v26`.

Performance wise, realistically speaking, I wouldn't be surprised that a pure software solution for encoding (and transcoding with [the `convert.html` benchmark](https://github.com/Vanilagy/mediabunny/pull/94#issuecomment-3257024385)) ends up faster. The code in PR is a bit faster than the previous benchmark also. All browsers provide vp8/vp9 encoding through software `libvpx`, so the browser has to readback the frame to CPU memory for encoding internally anyhow.
But it would definitely be faster for merging alpha + rendering the merge frame by keeping it all GPU if available (but GPU decoding is not supported in Firefox yet).

Vanilagy:
> While it is definitely a non concern for Mediabunny, it allows for separately publishing the alpha wrappers standalone for anyone to integrate (potentially) seamlessly.

It is planned for a near-future version of Mediabunny to expose custom decoder and encoder classes, so basically a "better" version of what WebCodecs offers, including the alpha handling. It shouldn't be too hard to adjust the current implementations for when alpha support lands in browser natively.

> From my understanding, the alpha and main must have keyframes at the same block otherwise seeking will result in decoder failure, at least in all native implementation.

Yeah, this is true, although I have found videos in the wild that don't care about this. I handle it by treating the pictures as opaque until a valid alpha key frame comes in.

> Be careful that browsers have [a hard limit (as low as 8 across all window) on the total WebGL context counts](https://stackoverflow.com/questions/61277222/how-to-determine-number-of-active-webgl-contexts). I personally would try to use a WebGL singleton to reduce chance of interfering with the users' WebGL application when multiple VideoDecoderWrapper / VideoEncoderWrapper are created. The users will certainly come to you when they find their WebGL application running into context lost.

Yeah, great point, I actually ran into that myself. Now, I do clean up the contexts myself using loseContext() (which fixed it for me), but I did worry about what happens if the user keeps many of these decoders/encoders around, which they are technically allowed to do. Thing is, I couldn't think of a clean solution. Having a singleton WebGL context to handle all of the GPU stuff is the obvious answer, but I would need to constantly resize the output framebuffer (canvas) as each decoder/encoder might have different dimensions that it works with. Not sure what the overhead there would be. You got any suggestions here?

> You should probably [gl.readPixels in a non blocking way](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_non-blocking_async_data_readback) or https://github.com/mrdoob/three.js/pull/28291.

I did try this today, although I found it makes the code around 2x to 3x slower, sadly, due to the polling logic involved. I think it's generally a cursed problem because ideally, we want to have maximum pipelining here, but when we wait for things like `gl.SYNC_GPU_COMMANDS_COMPLETE` we prevent pipelining. But then again, blocking on `readPixels` also blocks pipelining in a way since the main thread can't pump in more frames. Perhaps the WebGL logic should happen offthread, or WebGPU should be used instead (which I intentionally avoided due to thin support still).

The performance with `readPixels` is fine to me, for now. There are enough pauses in between calls to leave plenty time for the UI to update, and calls took around 2-4ms for 1080p frames on my Windows machine, which is still much faster than any in-JavaScript RGB to YUV conversion (which also blocks).

> You could probably omit the alpha shader by texImage2D with format gl.ALPHA and get the full Y plane already and then uint8.fill the UV after readPixels (similarly for merging, could use the original gl.LUMINANCE value without converting to RGB by browser internally).

I don't quite follow. I initially attempted something similar: Call `readPixels` only using `gl.ALPHA`, and then pad the rest of the buffer by calling `.fill()`, which seemed like the most efficient solution to me. Thing is, only `gl.RGBA` is supported in practice by browsers. But this relates to `readPixels` and not `texImage2D`, so could you explain what you mean?

> The recreating the VideoFrame with opaque format for "color" is only needed for a browser bug in Safari for 'I420A', might also be already fixed in latest version.

I tried this out and didn't notice any performance differences to be honest. Will keep it in the naive code for now, might do this optimization later if it proves effective.

---

Gonna ship this now and see how it plays out!

ghost:
> but I would need to constantly resize the output framebuffer (canvas) as each decoder/encoder might have different dimensions that it works with

If resizing is concerning, use a big enough, like `4096x4096` by default and only `readPixels` for the needed area. Need to workaround the problem that the size does not match in shader/texture sampler.

But I think it is fine to resize with `OffscreenCanvas`. The GPU need to internally switch to the [drawing buffer](https://registry.khronos.org/webgl/specs/latest/1.0/#THE_DRAWING_BUFFER) when executing the instructions. Size changing should be efficient with `drawingBufferStorage`.

Technically, you might need `drawingBufferStorage` for combing frames with HDR. But `VideoFrame` with HDR usage is is probably rare. I can only think of the non-standard `MediaStreamTrackGenerator` to display in `<video>` and [`<canvas>` but support is limited](https://www.github.com/whatwg/html/issues/9461#issuecomment-2777580077).

I think the main problem of resizing is with `HTMLCanvasElement` + `CanvasRenderingContext2D` (where states are reset on assignment), with all the layout recalculation goodness. That said, the performance is likely GPU driver dependant.

> I did try this today, although I found it makes the code around 2x to 3x slower, sadly

Well, because with `fenceSync`, it has to not interfere with any UI and GPU stuff going on. There is (usually) only 1 GPU after all. It is probably only helpful for encoding performance if it is implemented with concurrency. Sorry for wasting your time.

> calls took around 2-4ms for 1080p frames on my Windows machine, which is still much faster than any in-JavaScript RGB to YUV conversion (which also blocks).

I am talking about copying the RGB"A" to Y plane from JS. The step to extra and create `'I420'` for alpha image. I was benchmarking on macOS only on the conversion example:
```
let i = 0;
for (; i < byteLength / 4; i++) {
	dst[i] = srcShifted[i * 4];
}
dst.fill(255, i); // The `readPixels` implementation already filled it. To be fair.
```
```
1280x720 , avg: 2.50ms. `readPixels` avg: 4.25ms.
1920x1080, avg: 5.56ms. `readPixels` avg: 7.55ms.
3840x2160, avg: 26.3ms. `readPixels` avg: 16.9ms. // Well, 4k
```
But yea, the difference on amount of blocking time is not big. I would believe that the `readPixels` time would get worse on a real graphics application. It would be great if it does not block at all. The best way I can think of is `mapAsync` from WebGPU with a single R channel texture. But again, probably only performant with concurrency! So, it might not help your implementation.

> But this relates to readPixels and not texImage2D, so could you explain what you mean?

You could also set texture format for `texImage2D` and only copy that channel. https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D#type Performance wise, it is probably not that important since GPU is fast at converting YUV to RGB. 
My only concern when merging alpha: the RGB value might be rounded unexpectedly vs original `LUMINANCE` probably fine with UV (128,128) if the encoder works properly. To be "correct", we need to discard whatever UV values and only use the Y value for alpha.
- PS: I only took a rough look: seems like you are always using `gl.RGBA`.

ghost:
Anyways, great to get it shipped first!

I am definitely a newbie on this topic also.
Sadly, whatever I say might not really help on the current implementation. I should have realized this earlier!
Really need to try and figure out even if the theory is agreeable.

Vanilagy:
Hey dude, I'm not sure what prompted you to delete your GitHub account, but I hope you'll come back! I was still looking forward to continuing this conversation and collaborating with you on future topics, since you know your stuff. I hardly knew ye but ye will be missed! ❤️ 

---

I need to run some tests regarding the performance of resizing the thing. Perhaps that's the way, just keep a single context backed by an OffscreenCanvas around and resize that as needed. If the user is doing simple usage patterns, then typically the size would remain constant anyway.

Very interesting that your JS implementation was beating the GPU + readPixels implementation, it wasn't like that on my end. It was almost unacceptably slow (something 30ms), which prompted me to switch to a GPU approach. Don't think there's too much of a hardware difference (I even own an M4 Mac as well), maybe something about my implementation was fishy, but I couldn't really tell you why because in my mind it was just a single tight loop. The advantage of the JS approach is you can easily throw it on a worker and achieve concurrency that way, only minimally blocking the main thread. Although I'm actually not sure about the performance of `copyTo` in WebCodecs. It's sync, and might also need to perform readback if the frame is sitting in VRAM somewhere.

PR #124, opened 2025-09-17
Pin github actions to commit SHAs

studnitz:
This reduces the attack surface of supply-chain attacks and is considered a good practice.

Although in this case it's only Github-owned actions anyways, but better be safe than sorry!


Vanilagy:
Sure, thank you!

PR #126, opened 2025-09-18
fix(source): properly handle abort signals and Headers in UrlSource requests

aidenlx:
## Summary

This PR fixes issues with UrlSource not properly supporting `signal` and `Headers` instances from `requestInit`, and implements abortable URL sources in the process.

### 🔧 **Fixed RequestInit Handling**
- **Signal Support**: Fixed `mergeObjectsDeeply` breaking AbortSignal instances by replacing it with specialized `mergeRequestInit` function
- **Headers Support**: Proper handling of both `Headers` instances and plain objects in request merging
- **Signal Combination**: User-provided signals are now properly combined with internal abort controllers using `anyAbortSignal`

### 🛡️ **Implemented Disposable Interface** 
- Added `dispose()` method to abort all pending requests and clean up resources
- Added `[Symbol.dispose]()` for TypeScript using declarations support
- Prevents operations on disposed UrlSource instances

### ⚡ **Enhanced Retry Logic**
- `retriedFetch` now properly respects abort signals during retry delays
- Abort errors are no longer retried unnecessarily
- Proper cleanup of timeouts when aborted

### 🏗️ **Improved Worker Management**
- Workers now check disposal state and abort flags
- Abort controllers are properly cleaned up when workers are aborted
- Coordinated abort propagation across all workers

## Test Plan

- [x] Verify abort signals properly stop all network requests
- [x] Ensure Headers instances are preserved during request merging  
- [x] Test disposal effectiveness with multiple concurrent requests
- [x] Confirm retry logic respects abort signals
- [x] Validate user-provided signals are combined correctly

## Issues Fixed

- UrlSource not respecting `requestInit.signal` from user options
- `Headers` instances being converted to plain objects, breaking functionality
- `mergeObjectsDeeply` destroying AbortSignal instances
- No way to globally abort all pending UrlSource requests
- Retry logic ignoring abort signals during delays

This implementation provides robust abort signal handling that allows users to cleanly stop all UrlSource operations when needed, whether through explicit disposal or external abort signals.

### DIFF
--- a/src/misc.ts
+++ b/src/misc.ts
@@ -541,6 +541,95 @@ export const mergeObjectsDeeply = <T extends object, S extends object>(a: T, b: 
 	return result;
 };
 
+/**
+ * Returns an AbortSignal that aborts when any of the provided signals abort. If no signals are provided,
+ * returns undefined. If a signal is already aborted, the returned signal will be aborted immediately.
+ */
+export const anyAbortSignal = (signals: Array<AbortSignal | null | undefined>): AbortSignal => {
+	const validSignals = signals.filter((v): v is AbortSignal => v !== null && v !== undefined);
+
+	if ("any" in (AbortSignal as any)) {
+		return AbortSignal.any(validSignals);
+	}
+
+	if (validSignals.length === 0) return new AbortSignal()
+
+	// If there's exactly one, return it directly
+	if (validSignals.length === 1) return validSignals[0]!;
+
+	const controller = new AbortController();
+
+	const abortWithReason = (reason: unknown) => {
+		if (!controller.signal.aborted) {
+			const abortFn: unknown = (controller as unknown as { abort: (r?: unknown) => void }).abort;
+			if (typeof abortFn === 'function' && abortFn.length >= 1) {
+				(abortFn as (r?: unknown) => void).call(controller, reason);
+			} else {
+				controller.abort();
+			}
+		}
+	};
+
+	// If any is already aborted, abort immediately with its reason
+	for (const signal of validSignals) {
+		if (signal.aborted) {
+			const abortedReason = (signal as unknown as { reason?: unknown }).reason;
+			abortWithReason(abortedReason);
+			break;
+		}
+	}
+
+	if (!controller.signal.aborted) {
+		for (const signal of validSignals) {
+			signal.addEventListener('abort', () => {
+				const abortedReason = (signal as unknown as { reason?: unknown }).reason;
+				abortWithReason(abortedReason);
+			}, { once: true });
+		}
+	}
+
+	return controller.signal;
+};
+
+/**
+ * Merges RequestInit objects properly, handling AbortSignal combination and Headers objects.
+ * @internal
+ */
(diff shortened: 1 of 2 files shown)


Vanilagy:
Bit too much going on in this PR and it feels very AI-generated and thus lifeless, therefore I will be closing this.

Not everything from this PR I want to include in Mediabunny. However, some things are good points:
- Proper merging of `Headers` instances
- Abortable sources (this is a planned feature anyway)

I think I will implement these myself.

The ability to pass your own AbortControllers is something I don't want to allow. Aborting is left to the library, and the user is not supposed to fiddle with that. Yes, there should be a global .dispose() method on the source or the input, but not a per-request abort situation.

aidenlx:
Good to hear that disposable sources are on the schedule! Currently it defaults to retry forever, this way, some unrecoverable error would cause resource leaking, considering there is no way to dispose it externally. I would personally suggest the design and logic in Deno [@std/async retry](https://jsr.io/@std/async/1.0.14/unstable_retry.ts) function. 

The description of PR is indeed ai generated but the code itself is hand written with ai reviews, mostly for double checking where disposal were needed. The work is quite boring to be honest, so it's okay to consider the code to be "lifeless" :)

The most serious issue is about request init handling, I wonder if it's really necessary to use deep merging when there are only two props overridden, and a kindly suggestion, if external signal for request init is not supported, maybe considering Omit<RequestInit, "signal"> to explicitly exclude it from type def. 

Anyway, media bunny is indeed a great library. when drafting the PR, I can tell that your code is pretty sound and future proof, to be honest. Hope that the misunderstanding won't hinder my future contributions. Thanks again for this comprehensive library :)

aidenlx:
btw, I've implemented disposable api per [ECMAScript Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management) recently supported by node, chrome and Firefox. It would be a nice to have if abortable api could follow the spec :)

Vanilagy:
Check my commit, I've already added improved merging! And you can use the `getRetryDelay` function to manually specify the retry logic already.

And yes of course, abortable stuff with obviously implement Disposable, that's how it has to be done!

aidenlx:
Didn’t saw it when writing reply 😂 super fast and great work! 🎉 

btw, I’m already using getRetryDelay now as a workaround for disposing, I mention that api just as some kind of reference for delay api design improvement, not a must to have :) 

anyway, looking forward to see abortable in action

Vanilagy:
Oh yeah sure, but the API I provide with `getRetryDelay` is general enough that you could use it to mimic the behavior of `@std/async retry`. Just not out of the box of course, but you can achieve the same behavior, which is important.

I guess a possible utility is a function that you can call with the parameters from Deno, and it returns a function that you can pass to `getRetryDelay` which behaves accordingly. If you want, you can add a PR for that. Something like `createRetryDelayFunction`

Vanilagy:
Just added the dispose thing as well, check the attached commit. You can't dispose per-source, rather you dispose the input and it will clean up the rest. After you dispose the input, you cannot use it anymore.

PR #147, opened 2025-09-25
Improve discarded tracks error message when trying to convert to MP3

samohovets:
It took me around 20 minutes to remember that there's a WASM MP3 encoder by Mediabunny.

A better error message would’ve saved me some time 😃

@Vanilagy feel free to modify the approach - I’ve added an explicit check for MP3 since the encoder is only available as an installable module for now.


samohovets:
Oops, let me clean up the history

PR #148, opened 2025-09-25
Improve discarded tracks error message when trying to convert to MP3

samohovets:
It took me around 20 minutes to remember that there's a WASM MP3 encoder by Mediabunny.

A better error message would’ve saved me some time 😃

@Vanilagy feel free to modify the approach - I’ve added an explicit check for MP3 since the encoder is only available as an installable module for now.


Vanilagy:
Yeah, the errors when you hit this case are horrible. Been wanting to fix this for a while!

You can see how I did it in my commit, took a bit more work. The big thing I added was an `isValid` field, which is `false` if the required output track counts can't be reached. Since this was the error before "MP3 requires at least one audio track" which is a horrible error message. Now, the error looks like this:

image

samohovets:
Also wanted to say about this! It was a bit confusing to see the warning before the error, that was not related to the warning at all. 

Great job!

PR #158, opened 2025-09-30
Stop retrying UrlSource if CORS error is suspected

JonnyBurger:


Vanilagy (approved):
Good stuff, let's see how this performs in practice

tsekiguchi:
Hi! I've been running into some problems with this CORS check. 

In my electron application, I'm using a URL in URLSource that has been generated from my CDN to extract key frames using Mediabunny. If / when the fetch fails, mediabunny mistakenly thinks that this is a CORS error because my `window.location.origin` is `localhost:5173` in dev or `app://` in prod, but the source origin of my CDN url is `http://media.company.com/`.

Of course, I'm able to bypass this by setting my own getRetryFn, so I'm able to get it working for now.

However, I did have a couple thoughts:

1) Is there a more surefire way to tell if this is a CORS error to prevent this unintentional error? It seems like only comparing the `originOfSrc` to `window.location.origin` is too narrow of a gate.
2) Should Mediabunny be responsible for CORS checks?
3) If you would like to keep this in place, could we add a console.warn detailing that it failed because of suspected CORS error? And further, it might be nice to point to the getRetryFn as an override. Right now, the error only comes in as 'Failed to fetch' without much indication as to why.

Personally, I would remove this check, and keep the infinite backoff as is. getRetryFn is a great fallback for more customized behavior on an app to app basis. It feels more in the developers overall responsibility of the project to manage their own CORS, and without a way to list the current CORS, I suspect this is going to cause more errors than it prevents. However, I'd love to hear your guys' thoughts!

Vanilagy:
Hi, thanks for bringing this up. Coming to think of it, I agree the currently handling is insufficient; the signal of "the origin is different" is not enough, since lots of people use dedicated CDN origins to serve their media. The reason we used that signal is because it's basically the closest one to detecting CORS that we found; there's no direct way to detect it, and that's for good security reasons.

When I originally created UrlSource, it defaulted to always doing an infinite backoff no matter what. Reason being, you made a request which you expect to resolve, and since the network can be flakey sometimes, it's crucial that it be retried until the request can be fulfilled. Note that this would also indefinitely retry CORS-protected resources though, which is kinda weird.

@JonnyBurger had a more "one and done" use case for Mediabunny and now recommends [no retry at all](https://www.remotion.dev/docs/mediabunny/metadata) in his case, which is metadata extraction.

The CORS change was started because of this conversation here:
image

So, part of the difficulty here, in my eyes, is finding a reasonable default. I went with the default of infinite retry because while people develop in their office with gigabit speeds, they often forget that the internet can be extremely flaky out there in wild, especially on mobile or while traveling. I wanted to ensure that Mediabunny would eventually recover in these cases instead of the app breaking (of course, you can catch this error case too, but I doubt most people would think of that). This is especially relevant for players, which users expect to recover when they come back online.

Another obvious default that comes to mind is "no retry". The idea here is that the users will be forced to face the harsh truth of networks themselves, and then the error message would be annotated with a "You can tell Mediabunny to retry requests by doing..." message, hinting them to the API. The docs would specify it too, of course. This case *probably* caters better to the individual app like you said because it forces them to think, "Do I want to retry or not? Is this error out of my control or can I fix it?". With that, I could imagine Mediabunny providing utility functions to generate specific types of retry functions easily, for example the infinite backoff one.

A third option is to require the `getRetryDelay` parameter when the UrlSource is created. This forces the user to consciously think about the problem (and makes it impossible to forget), but also makes the API slightly harder to use - there's no default case they can opt into. I personally don't mind API designs like this - I prefer strictness and clarity over brevity - but I'm not sure about everybody else.

These are some of my thoughts! I would definitely agree with you that it may be out of scope for Mediabunny to handle CORS errors; it's the responsibility of the developer using Mediabunny to provide a URL that works. I also agree that the console errors and warnings should point to the `getRetryDelay` option as a means for configuration.

tsekiguchi:
Thanks @Vanilagy ! 

These are all very valid points. Infinite retry might lock a process, while no retry might throw too easily. For example, I tend to sometimes get 1 or 2 fetch fails from my CDN, especially when working with 100k+ files, so I personally wouldn't want to throw the entire process from one rogue fail.

Maybe a middle ground for all of this is removing the CORS check, and providing a reasonable default of 5 retries with exponential backoff. If the final one throws, then we can provide a short and sweet console.warning like:

`URLSource fetch failed after 5 attempts. Please check your CORS settings to ensure that the origin source is allowed. To increase retries, please modify the getRetryFn() callback in URLSource.`

If you agree, I'm more than happy to make this change in a separate PR

Vanilagy:
I think that would be a decent default, actually. One could extend this default to not retry at all if `navigator.onLine` is `false` (as that indicates no internet). The thing is that different defaults appear better depending on the use case. If a user could enter any URL, then it should fail fast due to CORS issues; but if the URLs are tightly controlled by the app itself, then that same logic doesn't apply anymore.

JonnyBurger:
Agree that this is painful and not surprised the wacky check had an edge case. 

For v1, I suggest we try to improve the heuristic to handle the Electron case. If you want you can task me with it, I'll try to come up with something next week.

My preference for v2 would still be that the default is to fail loudly. It makes more sense for most use cases, and the Media Player use case anyway requires thoughtful putting together of various lower-level Mediabunny APIs, a sample retry logic could be added to the Media Player example code, making the users also explicilty aware of the behavior.

Vanilagy:
I would say the case is not only limited to Electron, but generally any web app which serves media assets through a CDN on another origin - not uncommon. But I agree with Jonny that we should see if we can improve the heuristic at least somewhat, if possible.

I also agree with the preference for v2. Do the simplest, most unopinionated thing by default which is to fail, and then point the user to the `getRetryDelay` function as a way to customize the behavior. In that case, I would also provide a utility function that *generates* a retry delay function, with simple options such as retry count, max delay, backoff logic, etc.

---

Writing this, I had another idea: The thing about CORS errors that differentiates it from other network errors is that it can be reliably hit: meaning, if you know a resource is CORS-protected, you only need a single request to check that. Why is this relevant? Because you can now use the *first* request as a "CORS-check": If the first request succeeds, then you **know** that any requests after that failing cannot possibly be due to CORS. So CORS errors are completely ruled out if the first (or any) request has gone through. CORS is *only* a possibility if the first request failed: now the possibilities are either CORS, or any other network error.

You could use this to make the retry logic smarter: cap the retries to the first request, making sure CORS-protected resources throw quickly. All other requests after the first successful one can then have infinite retry, as the client can be pretty sure that the resource is indeed reachable.

tsekiguchi:
Agreed with v2. It should fail loudly and quickly, directing the user to implement their own retry fn if needed. I don't think any specific heuristics are needed for electron - this is really the only issue that I've run into so far. It really just boiled down to the problem of comparing the URLSource to the window location origin on error. As long as that check has been removed, then the rest is okay.

Beyond that, while additional CORS checks might be nice to haves, personally I think it might start to go beyond media bunny's purview.

Issue #159, opened 2025-09-30
set specific fragment durations in IsobmffOutputFormatOptions


AJFunk:
I see that `IsobmffOutputFormatOptions` provides a `minimumFragmentDuration` already, but this results in fragments that have durations > 1 second (seems to always be ~5 seconds). I need a way to to specify that my `Mp4OutputFormat` has fragment durations of exactly 1 second.

My situation is similar to https://github.com/Vanilagy/mediabunny/issues/103 

Is this possible already? I can't find any way to achieve this with the current implementation. Perhaps allowing the user to set a `maximumFragmentDuration` would solve this?

AJFunk:
I've implemented this in https://github.com/Vanilagy/mediabunny/pull/160 and based on testing within my own app, this seems to do the trick!

Setting both `minimumFragmentDuration = 1` and `maximumFragmentDuration = 1` I'm able to produce 1 second fragments

Vanilagy:
Have you tried setting `keyFrameInterval` to 1 in the encoding options? https://mediabunny.dev/api/VideoEncodingConfig#keyframeinterval

That's the much more kosher, proper way to solve this, as it means you'll actually produce proper keyframe-delimited fragments.

AJFunk:
> Have you tried setting `keyFrameInterval` to 1 in the encoding options?

Yes, I tried this but it doesn't seem to make a difference in the fragment duration (they're still 5 seconds)

Vanilagy:
How could that be? Hard to believe. The 5 seconds must come from somewhere. Can you show me your code?

Vanilagy:
Ah I apologize, I assume you're using the Conversion API, which `keyFrameInterval` is not available yet

AJFunk:
Ah ok I thought I was going crazy! Haha yes I am using the Conversion API and was trying to figure out how to pass the `keyFrameInterval` into either the `output` or the `video` (wasn't sure if I was using these wrong or something).

Is this something you plan to add support for? It seems like I could accomplish the 1 second keyframe intervals/fragment durations by implementing this myself (e.g. use VideoSampleSource and processing the frames myself) but the Conversion API is super convenient and this is currently the only missing piece for me (and it's a pretty urgent blocker for me at the moment)

Vanilagy:
[...]

PR #160, opened 2025-10-01
add maximumFragmentDuration to IsobmffOutputFormatOptions

AJFunk:
Adds `maximumFragmentDuration` to the `IsobmffOutputFormatOptions`. When used in conjunction with `minimumFragmentDuration`, it allows specifying a a range (e.g. 2-5 seconds) or a specific fragment duration (e.g. 1 second)

This addresses https://github.com/Vanilagy/mediabunny/issues/159

### DIFF
--- a/docs/guide/output-formats.md
+++ b/docs/guide/output-formats.md
@@ -64,6 +64,7 @@ The following options are available:
 type IsobmffOutputFormatOptions = {
 	fastStart?: false | 'in-memory' | 'reserve' | 'fragmented';
 	minimumFragmentDuration?: number;
+	maximumFragmentDuration?: number;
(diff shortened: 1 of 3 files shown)


AJFunk:
superseded by https://github.com/Vanilagy/mediabunny/pull/162

Vanilagy:
For the record, I still think a field like this would be good to have. The main usage being that some users may want to create a steady stream of fragments regardless of key frame cadence. I'm not sure if this could mess things up if B-frames are in the mix, however. Since B-frames can have wildly out of order timestamps, you might have the next fragment starting *earlier* than the previous one ended, which is just fucky. But putting a bounds on B-frames is non-trivial

PR #181, opened 2025-10-16
Fix the transfer characteristics map key for BT.2100 PQ

thurinus:
Fix for error:
```
Error during decodability check: TypeError: Failed to execute 'isConfigSupported' on 'VideoDecoder': Failed to read the 'colorSpace' property from 'VideoDecoderConfig': Failed to read the 'transfer' property from 'VideoColorSpaceInit': The provided value 'pg' is not a valid enum value of type VideoTransferCharacteristics.
    at InputVideoTrack.canDecode (base-DDKcIYxw.js:5655:42)
```

WebCodecs spec:
https://www.w3.org/TR/webcodecs/#videotransfercharacteristics

Confirmed fix against examples.


Vanilagy:
Thank you for catching this!

PR #186, opened 2025-10-18
Make Mediabunny more resilient against `data` boxes ending abruptly

### LINKED ISSUE #185: "Tried reading [3514, 3518), but slice is [3461, 3515)." when reading MP4 file
Mediabunny 1.24.0.

bug.ts
```ts
import {ALL_FORMATS, FilePathSource, Input} from 'mediabunny';

await new Input({
	formats: ALL_FORMATS,
	source: new FilePathSource('file.mp4'),
}).getPrimaryVideoTrack();
```

file.mp4:

https://github.com/user-attachments/assets/9579194f-8c49-4fdc-b26b-84d7b2f59266

```
bun bug.ts
```

I volunteer to try to debug and fix the issue.

JonnyBurger:
Fixes #185

Seems like videos from Getty Images are sometimes encoded like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,3 @@
-> [!NOTE]
-> I'm on vacation until 18 October, so expect slow or no replies to issues during that time. 🏖️
-
 # Mediabunny - JavaScript media toolkit
 
 [![](https://img.shields.io/npm/v/mediabunny)](https://www.npmjs.com/package/mediabunny)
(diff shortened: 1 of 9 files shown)


Vanilagy (approved):
Thanks for the fix. Definitely cursed that files are being written incorrectly. This is technically a bug that's in more places: if a box is shorter than its size says (or than it needs to be per-spec), the parser fails. I think this is fine, since the parser should fail for invalid files, the question is just the level of tolerance. For example, the data box reading code could still fail if the slice is shorter than the length field in the box header. Mediabunny generally isn't very resillient against this. Something for the future, maybe.

PR #187, opened 2025-10-18
Explain how to use examples dev server

JonnyBurger:
I struggled again to start the examples server.

- I guess this is because it prints `http://localhost:5173`, but it gives a 404
- `http://localhost:5173/examples/media-player` also gives a 404, you have to know to add index.html
- I also tried `cd examples && npm run dev` and `cd examples/media-player && npm run dev` and this starts a Vite server as well, but this is the wrong way of doing it.

Maybe I am an idiot, but I figure it doesn't hurt to be explicit!

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ npm run lint # ESLint
 
 npm run docs:generate # Generates API docs
 npm run docs:dev # Start docs development server
-npm run dev # Start examples development server, will run at http://localhost:5173/examples/[name]/index.html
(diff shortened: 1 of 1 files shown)


Vanilagy:
The right way to do is to append a final slash:

`http://localhost:5173/examples/media-player/`

I'm not sure why, but it needs it. But I agree, the repo should say that. Can you update the PR?

JonnyBurger:
@Vanilagy Done!

PR #190, opened 2025-10-20
feat: Skip unnecessary canvas clearing and drawing functions when not…

monsterooo:
Added per-frame drawing check to reduce rendering time by a few milliseconds; may have minimal impact during single video playback.


Vanilagy:
Not needed. The function only draws to the canvas when it switches to the next frame.

Issue #197, opened 2025-10-27
Unhandled error when disposing input immediately


JonnyBurger:
Repro to run: https://github.com/JonnyBurger/mediabunny-repro

The following code, when ran in a vanilla Vite app:

```ts
import { ALL_FORMATS, Input, UrlSource } from "mediabunny";

const input = new Input({
  source: new UrlSource("https://remotion.media/video.mp4"),
  formats: ALL_FORMATS,
});

input
  .computeDuration()
  .then(() => {})
  .catch(() => {});

input.dispose();
```

Results in an unhandled rejection that is logged to the console:

```
mediabunny.js?v=8a062a54:14897 Uncaught (in promise) Error: Response stream reader stopped unexpectedly before all requested data was read.
    at UrlSource._runWorker (mediabunny.js?v=8a062a54:14897:19)
	_runWorker @ mediabunny.js?v=8a062a54:14897
	await in _runWorker
	runWorker @ mediabunny.js?v=8a062a54:15426
	_retrieveSize @ mediabunny.js?v=8a062a54:14835
	await in _retrieveSize
	getSizeOrNull @ mediabunny.js?v=8a062a54:14628
	(anonymous) @ mediabunny.js?v=8a062a54:15601
	_getDemuxer @ mediabunny.js?v=8a062a54:15610
	computeDuration @ mediabunny.js?v=8a062a54:15634
	(anonymous) @ main.ts:9
```

<img width="779" height="258" alt="Image" src="https://github.com/user-attachments/assets/9402f66f-f18c-4a55-b71b-09eb24c51995" />

I'm puzzled actually. I could not reproduce this in the internal Mediabunny testbed, only when using the version from NPM!

I also tried:
- Disabling Vite module optimization
- Rewriting the alias to use `'./dist/modules/src/index.js'` instead (same as would be used if imported from NPM)

Vanilagy:
The fix was a bit more complicated. The issue was that workers were being created after the source was disposed (due to async stuff), and these workers weren't shutting themselves down. Simply suppressing the error like you did in your PR would've only been a duct-tape solution; the real solution is making the workers shut themselves down (and therefore avoid throwing the error) even in this scenario.

Will release this with the next version. Good find!

PR #203, opened 2025-10-29
Don't re-throw reader error when ReadOrchestrator is disposed

JonnyBurger:
This solves issue #197 :)

Tested: Without the change, we get an unhandled rejection when switching compositions in Remotion, after the change, it is gone.


Vanilagy:
Fixed it in the issue.

PR #210, opened 2025-11-05
Increase tolerance of pixels by 1 in tests

JonnyBurger:
This makes the test suite pass on macOS as well. Would be nice to get everything green


Vanilagy:
Flake heaven

PR #212, opened 2025-11-06
Preserve rotation when calling `.toVideoFrame()`

JonnyBurger:
Original customer issue: https://github.com/remotion-dev/remotion/issues/5916

I tested this by dragging the video into remotion.dev/convert.  
The media player, which uses a `CanvasSink`, correctly rotates the video.  
The thumbnail, which uses `VideoSampleSink`, does not do any rotation - which is okay but `rotation` is not set in `VideoFrameInit`, which leads it to be `0`. This results in the thumbnail being rotated, and in Remotion renders where we have to transfer the frame across tabs and we cannot use a CanvasSink for this reason, the video also rotates.

`VideoFrameInit -> rotation` is not yet in the types but in the spec: https://www.w3.org/TR/webcodecs/#dom-videoframeinit-rotation

I tested the fix by changing the `node_modules` of remotion.dev/convert and confirmed this fixed the issue.

### DIFF
--- a/dev/convert.html
+++ b/dev/convert.html
@@ -48,6 +48,7 @@
 			target
 		});
 
+		let ctx = null;
 		const conversion = await Mediabunny.Conversion.init({
 			input: new Mediabunny.Input({
 				formats: Mediabunny.ALL_FORMATS,	
@@ -56,6 +57,7 @@
 			output,
 			audio: (_, n) => ({
 				discard: n > 1,
+				//codec: 'opus',
 				/*
 				process: (sample) => {
 					return sample;
@@ -97,7 +99,7 @@
 			},
 			*/
 			video: () => ({
-				/*
+				//discard: true,
 				process: (sample) => {
 					if (!ctx) {
 						// Create a canvas for image compositing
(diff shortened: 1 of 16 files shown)


Vanilagy:
I was aware of the `rotation` field but assumed it's still too early to use in practice because it hasn't been in the spec for so long. But seems like Chromium understands it well. I've also seen Safari use that field, actually.

To complete this PR, can you make it extract the rotation from the VideoFrame when a VideoSample is constructed from one? This way, the transformation to and from VideoSample is lossless.

---

Regarding usage: in my own projects, I typically use VideoSample everywhere and use its `.draw()` method (which respects rotation) rather than accessing the underlying `VideoFrame`. Too bad there's no way in JS to create your own [structured clonable type](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types); if there were, you could just send the VideoSample across tabs directly.

JonnyBurger:
Pushed!

---

We now also started using VideoSamples where possible (e.g. `extractFrames()` snippet).
Unfortunately yes it is not transferrable, so we'll be stuck with `VideoSampleSink`.

JonnyBurger:
Also PR'd to `@types/dom-webcodecs`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74027

Vanilagy:
I have made further modifications to the drawing functions on `VideoSample` so that they now behave correctly even when the underlying `VideoFrame` is rotated. Bit tricky this, but the tests should hopefully provide some stability here.

I'll be merging this PR into the v2 branch. This change is breaking, as code that previously expected an unrotated VideoFrame and handled rotation itself now receives a pre-rotated VideoFrame from Mediabunny. This change is still, but it's breaking.

PR #214, opened 2025-11-06
feature(isobmff): read disabled tracks

### LINKED ISSUE #213: Reading ISOBMFF with multiple audio tracks
Hi! And thank you for your work! I attended Demuxed last week and got introduced to your work by the talk there, very much appreciated.

I tried to load some multi audio track ISOBMFF files. They contain multiple audio tracks for different languages among some other things. I noticed that mediabunny did not read all of the tracks, seems like it skips tracks that do not have the enabled flag set: https://github.com/Vanilagy/mediabunny/blob/5aabf6350a28a57896f6ad67579692b21f6db6b0/src/isobmff/isobmff-demuxer.ts#L698

Would it be possible to remove this check to allow processing for all audio tracks, even if they are not enabled by default? Or is there some other way to process disabled audio tracks?

lideen:
Allows tracks to be parsed even if they are disabled.

Closes #213

### DIFF
--- a/src/isobmff/isobmff-demuxer.ts
+++ b/src/isobmff/isobmff-demuxer.ts
@@ -92,6 +92,7 @@ type InternalTrack = {
 	id: number;
 	demuxer: IsobmffDemuxer;
 	inputTrack: InputTrack | null;
+	isDefault: boolean;
 	timescale: number;
 	durationInMovieTimescale: number;
 	durationInMediaTimescale: number;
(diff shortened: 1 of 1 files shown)


Vanilagy (approved):
I say this is a fine addition given that this reflects FFmpeg's behavior. I did make some modiciations to ensure that enabled tracks are treated as "default", meaning they are sorted so they come before the non-enabled tracks. This ensures that calls to `getPrimaryVideoTrack()` etc. return the primary track, not the first track.

Issue #184, opened 2025-10-17
Video conversion crashes with large input files on devices with low memory


hexcowboy:
Recorded on iPhone Mini 12 running iOS.

https://github.com/user-attachments/assets/316be1b2-5413-45dc-8235-b132f9758b3e

https://github.com/user-attachments/assets/4a1ef258-e717-422a-bc73-b9d5f48896e2

The video uploaded was a 253MB Quicktime MOV.

```
[Error] Failed to load resource: WebKit encountered an internal error (e74a667e-907f-4f45-951b-fd912f7ea65a, line 0)
[Error] Failed to load resource: WebKit encountered an internal error (c9d700fc-4b54-4335-9f21-feed0db2e370, line 0)
[Error] Failed to load resource: WebKit encountered an internal error (3ff53954-14d5-49b3-bab4-6dab986722a3, line 0)
[Error] Failed to load resource: The operation couldn’t be completed. (WebKitBlobResource error 1.) (d1b6241e-52f3-4033-b6b3-7560d7df54ee, line 0)
[Error] NotFoundError: The object can not be found here.

	(anonymous function) (file-compression-9U2SBdhf.js:72)
[Warning] Conversion already canceled. (base-DDKcIYxw.js, line 19804)
```

I think the best way to reproduce is to set a large input file in any MediaBunny conversion. I used this page https://mediabunny.dev/examples/file-compression/

Vanilagy:
Fixed in #215

PR #215, opened 2025-11-08
Use `ArrayBuffer` instead of `ReadableStreamReader`on WebKit browsers

hexcowboy:
After investigating https://github.com/Vanilagy/mediabunny/issues/184, I found that the problem occurs in WebKit browsers on iOS, especially on devices with lower memory.

Instead of using a `ReadableStreamReader`, we can use `ArrayBuffer` which is much more stable on mobile WebKit browsers. I can't find any definite causes from WebKit source, but I've added some likely reasons in a comment in my changes. This change does not come at a performance cost from my own tests on desktop Safari, obviously doesn't affect other browsers, and it fixes page crashes on WebKit browsers (iOS Safari, iOS Chrome, etc) when uploading large media files.

### DIFF
--- a/src/media-sink.ts
+++ b/src/media-sink.ts
@@ -26,7 +26,7 @@ import {
 	insertSorted,
 	isFirefox,
 	isNumber,
-	isSafari,
+	isWebKit,
 	last,
 	mapAsyncGenerator,
 	promiseWithResolvers,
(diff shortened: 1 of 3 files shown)


Vanilagy:
Thanks for investigating this and providing a fix! I have simplified the `isWebKit` condition a lot, seems like checking `navigator.vendor` is all that is needed. Tell me if I missed anything here. I also got rid of `isSafari` because those checks really want to check for WebKit, not Safari.

You should make sure to report this bug to WebKit, though! Seems like the bug occurs when a reader (spawned from a Blob) is pulled over and over again on a large file, on an iPhone. I wasn't able to reproduce the bug on my Mac, for instance. So yes, while Mediabunny can work around the issue for now, we should make sure that WebKit already knows about (and link the report in the code) or report the bug ourselves. Can you do that?

hexcowboy:
I've reported the relevant WebKit bug here https://bugs.webkit.org/show_bug.cgi?id=302282

PR #245, opened 2025-12-10
Seek to time when pointer moves in player example

abouolia:
https://github.com/user-attachments/assets/f3a845ae-0f49-47c8-bf98-952bea45cdcb


Vanilagy:
This is cool and you're free to do this in your own video player based on Mediabunny, but I intentionally did not do this here. If the decoder is slow, more complex debouncing logic is needed to make this feature work correctly, and I didn't want to overcomplicate the player example.

PR #249, opened 2025-12-12
Add Dark Mode Remotion Logo in README

JonnyBurger:
The blue logo doesn't look good on dark background, so we have a dark mode version of it.

On the website, it doesn't seem to easy to implement without major hacks, so I'll left that as is 😄


Vanilagy (approved):
Didn't think the blue one looked that bad!!

PR #254, opened 2025-12-16
`Conversion.init(hardwareAcceleration)` option

### LINKED ISSUE #253: Conversion.init() doesn't accept specific `VideoEncoderConfig`
Follow-up of https://github.com/Vanilagy/mediabunny/issues/61 and https://github.com/Vanilagy/mediabunny/issues/222

I found to have unreliable results between `canEncodeVideo({hardwareAcceleration: true})`  and actual encoding results ¹

While `VideoEncoder.isConfigSupported({"codec":"avc1.640016","width":640,"height":480,"bitrate":1000000,"alpha":"discard","hardwareAcceleration":"prefer-hardware","avc":{"format":"avc"}})` was consistent with `canEncodeVideo({hardwareAcceleration: 'prefer-hardware'})`, 

is consistent with `await canEncodeVideo({hardwareAcceleration: 'prefer-hardware'})`

(actually `{supported: false}`)

Actual encoding test using ` await Conversion.init() /  execute()` succeeded.

The underlying confusion is that `Conversion` API takes a `ConversionVideoOptions` but no way to specify custom `VideoEncodingConfig` which is thus hardcoded :

https://github.com/Vanilagy/mediabunny/blob/ea9ba96ae3b7ef7a1835187ad61039ecf0414a6b/src/conversion.ts#L981-L987

without any specific `hardwareAcceleration`.

¹ I need to compare whether software-only Mediabunny is more efficient than `ffmpeg.wasm` which is why I'd like to assess the actual hw/sw specific performances.

drzraf:
Support `Conversion.init(hardwareAcceleration)` option passed down to the `VideoEncodingConfig`

fix #253

### DIFF
--- a/docs/guide/converting-media-files.md
+++ b/docs/guide/converting-media-files.md
@@ -124,9 +124,9 @@ type ConversionVideoOptions = {
 	crop?: { left: number; top: number; width: number; height: number };
 	frameRate?: number;
 	codec?: VideoCodec;
-	hardwareAcceleration?: 'no-preference' | 'prefer-hardware' | 'prefer-software';
 	bitrate?: number | Quality;
 	alpha?: 'discard' | 'keep'; // Defaults to 'discard'
+	hardwareAcceleration?: 'no-preference' | 'prefer-hardware' | 'prefer-software';
 	keyFrameInterval?: number;
 	forceTranscode?: boolean;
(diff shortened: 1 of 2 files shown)


Vanilagy:
will ship this in the next version

PR #255, opened 2025-12-16
Add failing `sample.allocationSize()` on iPhone video test case

samohovets:


Vanilagy:
Thanks, but I think I'll try to add a more synthetic test that reproduces this.

PR #261, opened 2025-12-29
Add support for 2d canvas merger workflow

lavrton:
On my tests, 2d canvas merger is 2-3x faster than webgl version when I run it inside puppeteer instance.
I don't know if it is possible to automatically detect which one is faster/better and use by default.
So maybe a user should control that.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -174,6 +174,7 @@ export {
 	VIDEO_SAMPLE_PIXEL_FORMATS,
 } from './sample';
 export {
+	AlphaMergerType,
 	AudioBufferSink,
 	AudioSampleSink,
 	BaseMediaSampleSink,
@@ -182,6 +183,7 @@ export {
 	EncodedPacketSink,
 	PacketRetrievalOptions,
 	VideoSampleSink,
+	VideoSampleSinkOptions,
 	WrappedAudioBuffer,
 	WrappedCanvas,
 } from './media-sink';
--- a/src/media-sink.ts
+++ b/src/media-sink.ts
@@ -824,7 +824,7 @@ class VideoDecoderWrapper extends DecoderWrapper<VideoSample> {
 	alphaHadKeyframe = false;
 	colorQueue: VideoFrame[] = [];
 	alphaQueue: (VideoFrame | null)[] = [];
-	merger: ColorAlphaMerger | null = null;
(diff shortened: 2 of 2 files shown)


Vanilagy:
There is likely a more direct way to perform this merging by using `VideoFrame.copyTo` instead of using canvases here. This was actually the original implementation of it, until I decided to use a GPU-based implementation instead.

In normal cases, it's also very performant. Here's it merging two 720p frames:
Screenshot 2025-12-30 at 14 43 05

It's probably not possible to reach this level of performance with a CPU-based algorithm - *on systems that have a GPU.*

But either way, having this be controlled by an option in the sink feels suboptimal to me, I don't know. This is an implementation detail, and Mediabunny should somehow find out which variant is more performant. The only way I could think of is that it finds it out experimentally; it tries both approaches and sticks with the one that was faster. But, kinda cursed, idk.

Closing this PR as I don't think it's the right solution. But if you want, feel free to explore a solution that uses `copyTo` instead. Especially when dealing with I420 frames, the merging is trivial, as it'll require a single memcpy operation. If we're in RGBA format, it'll be slower, as we'll need to do a strided scatter operation. Also doubt any SIMD will help here, so there's little reason to switch to WASM for this. And, optimally, when these operations are expensive, they should run in a separate thread.

lavrton:
Thanks for looking into that. I tried `VideoFrame.copyTo`. It gave me the same performance (no, or very little boost). While the code gets a bit more complex, because of async flow (as it effects the caller, etc). 

Maybe the name of the arguments can be different. E.g. instead of `alphaMerger: 'webgl' | '2d'` use something like `type AlphaMerge = "auto" | "gpu" | "cpu";`. So on auto, it will auto-detect as you suggested. While `gpu` will be a default. 

Or as an alternative, add ability to overwrite `alphaMerger` with a custom class:

```js
class ColorAlphaMerger {
  // implementation
}

const videoSink = new CanvasSink(videoTrack, {
  poolSize: 5,
  fit: 'contain',
  alpha: videoCanBeTransparent,
  alphaMerger: new ColorAlphaMerger();
});
```

We render on both server-side and client-side. Such a good performance improvement on the server-side is important.

Is there anything I can do from my side to have any implementation? Suggest another solution? Make another pull request? Make a testing repository where we can quickly see the performance on both desktop and puppeteer?

Vanilagy:
I'm still thinking about which solution would be best for this, will get back to you.

PR #273, opened 2026-01-11
fix(matroska): defer DefaultDuration calculation until TimestampScale is known

jepcd:
When decoding certain MKV files, an error is thrown:
```
TypeError: duration must be a non-negative number.
```

This happens because the DefaultDuration can appear before the TimestampScale in the file. When DefaultDuration is processed, `timestampFactor` is still `-1` (uninitialized), causing the calculation to produce a negative duration:
```
// timestampFactor = -1 (not yet set)
// DefaultDuration from file = 41708333 ns
defaultDuration = (-1 * 41708333) / 1e9 = -0.0417  // Negative!
```

I've deferred the calculation of defaultDuration until timestampFactor is known to be set. 

The file that was failing to play before now works as expected


Vanilagy (approved):

Vanilagy:
Looks good to me

PR #157, opened 2025-09-30
Add E-AC-3 (Dolby Digital Plus) and AC-3 (Dolby Digital) audio codec support

wiedymi:
Implements E-AC-3 and AC-3 audio codecs using FFmpeg for encoding/decoding.

Features:
- Full E-AC-3 (Enhanced AC-3) encoder and decoder
- AC-3 (Dolby Digital) encoder and decoder
- WASM-based implementation using FFmpeg's libavcodec
- Worker-based processing for non-blocking operation
- Support across all containers (AVI, MP4, MOV, MKV)

Package structure:
- packages/eac3/ - Standalone codec package
- Uses Emscripten-compiled FFmpeg for WebAssembly

Codec integration:
- Added 'ac3' and 'eac3' to NON_PCM_AUDIO_CODECS list
- Codec string mapping (ac-3 for ac3, ec-3 for eac3)
- Validation in validateAudioChunkMetadata

Container support:
- ISOBMFF (MP4/MOV): ac-3 and ec-3 sample entries
- Matroska: A_AC3 and A_EAC3 codec strings

Tests:
- Updated 4 test suites (AVI, MP4, MOV, MKV) to include ac3/eac3
- 6 new test sample files for ac3/eac3 across containers
- Total 12 additional tests validating ac3/eac3 codec support
@[wiedymi](https://github.com/Vanilagy/mediabunny/commits?author=wiedymi)
wiedymi committed 5 minutes ago

PR #275, opened 2026-01-12
feat(codec): add detection support for AC-3, E-AC-3 audio codecs

jepcd:
Adds support for detecting additional codecs within isobmff and matroska. 

This lets users provide their own encoder/decoder via the Custom coders API.

### DIFF
--- a/src/codec-data.ts
+++ b/src/codec-data.ts
@@ -2658,7 +2658,7 @@ export const AC3_SAMPLE_RATES = [48000, 44100, 32000] as const;
  */
 export const AC3_ACMOD_CHANNEL_COUNTS = [2, 1, 2, 3, 3, 4, 4, 5] as const;
 
-export interface AC3FrameInfo {
+export interface Ac3FrameInfo {
 	/** Sample rate code */
 	fscod: number;
 	/** Bitstream ID */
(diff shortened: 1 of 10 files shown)


Vanilagy:
Generally a good idea, however the exact meaning of these codecs (data format, codec string, description) should be explicitly defined somewhere in the docs, similar to the WebCodecs Registry, so there's definitely more work to do in this regard. Wanted to do a Mediabunny Codec Registry for a while now.

Also, the question is where the integration of these codecs in the corresponding container is spec'd. Which resources did you use? For Matroska, there is https://www.matroska.org/technical/codec_specs.html, although this doesn't contain `A_EAC3`, but FFmpeg does, so that's a good sign. Couldn't find anything on the ISOBMFF side though, maybe FFmpeg is the only real reference here.

jepcd:
Yep, still quite some work/research required on this so I put it up as a draft. Plan to work more on it.

Just wanted to see if you'd be open to merging this. I see [another large pr](https://github.com/Vanilagy/mediabunny/pull/157) that implements the extension too, but maybe implementing the detection would be a better first step.

Vanilagy:
Yup for sure, the detection is the first step. Note that you can already get packets for unknown codecs (discriminated using `internalCodecId`), but they don't plug into the existing decoder/encoder infrastructure for Mediabunny, that's right.

jepcd:
@Vanilagy I have removed DTS from this PR as it is already large enough.

The references used are specified in the source code, though will also mention them here.

The codec string (ac-3, ec-3) is mentioned here: https://mp4ra.org/registered-types/codecs

The main reference for my implementation was [ETSI TS 102 366 V1.4.1](https://www.etsi.org/deliver/etsi_ts/102300_102399/102366/01.04.01_60/ts_102366v010401p.pdf)
This covers integration into ISOMBFF (Annex F), and syncframe/bitstream layout which I had to parse to get the decoder config.

There are 2 tests for each format, and I also verified that the dec3 box created when remuxing from mkv to mp4 is identical to ffmpeg

I updated the docs to add ac3/eac3 but with a note that says you need to supply a custom coder. 

Let me know if you would like any changes!

jepcd:
Added ac3/eac3 support for mpeg-ts!

wiedymi:
Duplicated of https://github.com/Vanilagy/mediabunny/pull/157?

Vanilagy:
@wiedymi Not really, this one's more focused on just the demuxing/muxing part and it's more complete, I will review this one soon

jepcd:
Compared to the other PR. this one

- Does not implement a encoder/decoder (you can use custom encoder/decoders though!)
- Does not attempt to include AVI support

This is more complete on the muxing/demuxing side as it follows the spec closely:
- Supports reading/writing `AC3SpecificBox`/`EC3SpecificBox` for isobmff
- Supports reading ac3/eac3 in System-B (though not very widely used, its in the spec and supported by ffmpeg)
- Support for mpeg-ts

wiedymi:
Got it, then I need to rebase mine work once it merged 👀

AVI there because of stacked PRs.

Vanilagy (approved):
Made a few small changes, mainly moving code around and making some parts of the code more strict. Feel free to look through them to see if I messed up anything.

Otherwise, looks good! Thanks for this effort. I'll deploy this PR in the next version of Mediabunny.

The logical next step would be to add extension packages that add AC3 & EAC3 decoder/encoder support, I guess.

PR #276, opened 2026-01-12
fix(matroska): handle FlagEnabled element correctly regardless of order

jepcd:
When parsing certain MKV files with disabled tracks (`FlagEnabled=0`) issues occurred:

```
TypeError crash: TypeError: null is not an object (evaluating 'this.currentTrack.decodingInstructions')
```
The code didn't check if this.currentTrack was non-null. FlagEnabled could set `this.currentTrack = null`

Additionally FlagEnabled might be read before or after the inputTrack was set, so we handle both cases.

Let me know if you need a reproduction file.

### DIFF
--- a/src/matroska/matroska-demuxer.ts
+++ b/src/matroska/matroska-demuxer.ts
@@ -1014,6 +1014,9 @@ export class MatroskaDemuxer extends Demuxer {
 
 				this.readContiguousElements(slice.slice(dataStartPos, size));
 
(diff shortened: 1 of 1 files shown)


Vanilagy:
```ts
					// Mark the track as disabled. Element order within TrackEntry
					// is not guaranteed, so FlagEnabled could appear before or after the track has been added to
					// the tracks array. We need to handle both cases.
```
How? They always get added after the call to `this.readContiguousElements`, after which a return is added, so it's impossible

Issue #284, opened 2026-01-20
Trimming with Conversion on MP4 URL source hangs


fredrikj:
I am doing a quite simple task of trimming an MP4 file obtained from a URL. When doing this, it hangs at some place in the process. There is no error log. Things just stops. Sometimes after having gotten some way, sometimes even before starting.
When error happens even before starting, I see a pending request for the source MP4 in the browser's network log. If the error happens later, I see no clue in the network log.

I have not seen this problem if my trimming starts at time 0.

```ts
   const input = new Input({
      source: new UrlSource('https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4'),
      formats: ALL_FORMATS,
    });

    const target = new BufferTarget();
    const output = new Output({
      target,
      format: new Mp4OutputFormat({
        fastStart: 'in-memory',
      }),
    });

    const conversion = await Conversion.init({
      input,
      output,
      trim: {
        start: 103,
        end: 274,
      },
      video: {
        codec: 'avc',
      },
      audio: {
        codec: 'aac',
      },
    });
```

Vanilagy:
Can you share more details about the exact environment you are able to reproduce this in? I ran your code four times and it never hung. Running Chrome on Windows 11.

fredrikj:
I think this issue can be closed. I am running Chrome on MacOS, but having a hard time reproducing it now. One important fix we did was to add `dispose`:
```
input = ...
try {
 ...
} finally {
  input.dispose();
}
```
That was our mistake to not have. A further thing might have been that I kept looking in the Chrome network log. It shouldn't matter, but with the whole developer tools closed it seems to behave better. We are adding a GET entry to the log for each frame, with our way of adding graphics to a canvas, so we certainly have lots of network calls in the log.

Thank you for your time, and sorry for my noise!

Vanilagy:
It's alright. Thing is, disposing the input is actually optional. If you don't, it'll be disposed once GC hits it. The dispose function is just good for disposing it *now*. This shouldn't really effect the conversion hanging though.

If you say you're fine for now, then so it is. If you ever encounter this again though, please reopen this issue, so we can get to the bottom of this.

PR #289, opened 2026-01-26
Set max one worker for UrlSource

fredrikj:
In my React app, Chrome sometimes got stuck in a pending state for a partial request of an MP4 file. Setting maxWorkerCount to 1 solved the problem. This fixes issue #284

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -416,9 +416,7 @@ export class UrlSource extends Source {
 
 		this._orchestrator = new ReadOrchestrator({
 			maxCacheSize: options.maxCacheSize ?? (64 * 2 ** 20 /* 64 MiB */),
-			// Most files in the real-world have a single sequential access pattern, but having two in parallel can
-			// also happen
-			maxWorkerCount: 2,
+			maxWorkerCount: 1, // Chromium has been seen to have problem with more than one worker
 			runWorker: this._runWorker.bind(this),
 			prefetchProfile: PREFETCH_PROFILES.network,
 		});


Vanilagy:
UrlSource has 2 workers for parallelism intentionally. The use case is when data in a file is spread out (imagine all video data in one chunk, then all audio data in another chunk), two simultaneous requests (they are range requests, remember) can go read this data instead of having one after another.

Still want to get to the bottom of this though. It could be that the parallelism throws off some HTTP servers, that's my theory, I doubt it is a Chromium bug. Are you sure this is a Chromium bug? Have you been able to reproduce it in other browsers? What HTTP server are you using?

I *can* imagine a new `parallelism` option for `UrlSource` which the user can manually set to 1. What do you think of this?

fredrikj:
I have tried on Chromium (Chrome, Arc and Edge) and Safari. Our MP4 files are on AWS S3 behind Cloudfront. I could reproduce the problem on all variants of Chromium (Chrome and Arc in Mac OS, Edge in Windows) but never got the problem in Safari.

I could also reproduce the problem if within the same frontend code fetching https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4, though I didn't try Safari on that. 

It seems that even if this would be a backend problem, there is still a frontend need to avoid it. Attaching a screenshot where the BigBuckBunny URL has been hardcoded into our React frontend. Only way to get out of this "pending" state is to reload the page. An option to use only one worker would be much appreciated!

Screenshot 2026-01-27 at 17 53 33

Vanilagy:
Can you PR a `parallelism` option for UrlSource (which just maps to worker count under the hood)? Wouldn't mind merging that in.

fredrikj:
Ok, I created PR #292. I hope it looks ok!

Issue #309, opened 2026-02-24
add support for PASP box in mov/mp4


tonybart1337:
some videos are not displayed correctly because mediabunny doesn't parse pasp box

[mp4boxjs ref](https://github.com/gpac/mp4box.js/blob/main/src/boxes/pasp.ts)

Vanilagy:
Can you send a video that is displayed incorrectly?

tonybart1337:
```bash
ffmpeg -f lavfi -i testsrc2=duration=5:size=640x360:rate=30 \
  -c:v libx264 \
  -pix_fmt yuv420p \
  -vf "setsar=2/1" \
  -t 5 \
  sar_2x1.mp4
```

[sar_2x1.zip](https://github.com/user-attachments/files/25525950/sar_2x1.zip)

tonybart1337:
@Vanilagy would you consider a PR for this feature?

I'm migrating our production video tooling to mediabunny and it's currently a blocker for us

Vanilagy:
Support will ship with the next version! `InputVideoTrack` and `VideoSample` now expose `pixelAspectRatio`, `squarePixelWidth` and `squarePixelHeight`, and the samples are drawn correctly according to their PAR.

tonybart1337:
Amazing! Thank you for the update!

Vanilagy:
Released in v1.35.0! Please try it out and confirm that it works for you guys.

tonybart1337:
@Vanilagy it works great, including the case with non squared frames and rotation

Vanilagy:
Lovely!

PR #310, opened 2026-02-25
pixel aspect ration support

tonybart1337:
implements https://github.com/Vanilagy/mediabunny/issues/309

I added new fields:
```
hSpacing
vSpacing
```

for most containers they're just 1:1, but for mov/mp4 they're extracted from PASP box if present
I also passed these fields to the Sample and applied in CanvasSink - since the crop is defined in coded pixels space

I did run some tests using media player example using:
```
npm run docs:build
npm run docs:preview
```

and it works great
before:
image

after:
image

### DIFF
--- a/docs/guide/reading-media-files.md
+++ b/docs/guide/reading-media-files.md
@@ -220,13 +220,19 @@ In addition to the [common track metadata](#common-track-metadata), video tracks
 videoTrack.codedWidth; // => number
 videoTrack.codedHeight; // => number
 
-// Get the displayed pixel dimensions of the track's samples, after rotation:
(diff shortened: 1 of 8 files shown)


Vanilagy:
Thanks for the PR, but I already started work on this before you created it 😅 Solved in this commit:
https://github.com/Vanilagy/mediabunny/commit/bb4749cd8d6b6ef2474eb2a69cae784d7e9ccfed

The solution is more complicated than what's in this PR. Not only did read support have to be added to all demuxers that support video tracks, but **write** support was also added. This feature also had ripple effects on how `VideoSample`s behave and on the Conversion API.

PR #317, opened 2026-03-06
Fix CTS=0 in fragmented fMP4 with multiple tracks

kzahel:
## Summary

When muxing fragmented MP4 with both video and audio tracks, `compositionTimeOffset` (CTS) is zero for all video samples. This causes B-frame content to display in decode order instead of presentation order, producing visible judder (e.g. ~8fps on 24fps content).

- **Root cause**: During `finalize()`, `interleaveSamples(true)` triggers `finalizeFragment()` via the cross-track keyframe check in `addSampleToTrack`. This writes the `trun` box before `processTimestamps` has computed correct `decodeTimestamp` values, so CTS = PTS - DTS = 0.
- **Fix**: Call `processTimestamps()` for all tracks at the start of `finalizeFragment()`. This is safe because `processTimestamps` is a no-op when `timestampProcessingQueue` is empty.
- **Video-only fragmented muxing is unaffected** because the single-track case never triggers the cross-track keyframe check during `interleaveSamples`.

## Reproduction

Using the existing test fixture `test/public/video.mp4` (H.264 with `has_b_frames=2` + AAC audio):

1. Demux video + audio tracks
2. Re-mux as fragmented fMP4 using `Mp4OutputFormat({ fastStart: 'fragmented' })`
3. Read back the output and iterate video packets in decode order
4. **Bug**: All packet timestamps are monotonically increasing (CTS=0, so PTS == DTS)
5. **Expected**: Timestamps are non-monotonic in decode order (B-frame PTS reordering preserved)

## Test plan

- [x] New test added to `test/node/isobmff-muxer.test.ts`: "Fragmented fMP4 with video+audio preserves B-frame CTS"
- [x] Test **fails** without the fix (timestamps monotonically increasing = CTS lost)
- [x] Test **passes** with the fix (timestamps non-monotonic = B-frame reordering preserved)
- [x] All 66 existing node tests pass with no regressions

## Real-world impact

Confirmed this fix resolves visible playback judder on 24fps B-frame H.264 content served as fMP4 segments via HLS in production.

### DIFF
--- a/src/isobmff/isobmff-muxer.ts
+++ b/src/isobmff/isobmff-muxer.ts
@@ -1247,11 +1247,11 @@ export class IsobmffMuxer extends Muxer {
 
 		const trackData = this.trackDatas.find(x => x.track === track);
 		if (trackData) {
-			this.processTimestamps(trackData);
-
 			if (trackData.type === 'subtitle' && track.source._codec === 'webvtt') {
 				await this.processWebVTTCues(trackData, Infinity);
 			}
+
+			this.processTimestamps(trackData);
 		}
 
 		if (this.allTracksAreKnown()) {
@@ -1273,11 +1273,11 @@ export class IsobmffMuxer extends Muxer {
 		this.allTracksKnown.resolve();
 
 		for (const trackData of this.trackDatas) {
-			this.processTimestamps(trackData);
-
 			if (trackData.type === 'subtitle' && trackData.track.source._codec === 'webvtt') {
 				await this.processWebVTTCues(trackData, Infinity);
 			}
+
+			this.processTimestamps(trackData);
 		}
 
 		if (this.isFragmented) {


kzahel:
Sorry about the drive by PR especially a Claude produced one. The diff is very small though and fixes my use case. Feel free to close PR etc and pull in yourself etc, or whatever you want. Thanks so much for this project! It's really freaking cool! I was trying to use ffmpeg wasm to probe and mux/demux but it is DOG SLOW with mountfs. Mediabunny is AWESOME. Thank you.

Vanilagy:
Thanks for bringing this up! The "root cause" fix by Claude was indeed misguided, the actual problem was that the timestamps wouldn't be processed when a track was closed, meaning when the fragment was then later finalized (which is now a legal operation due to the track being closed), the timestamps were still unprocessed. The fix from Claude here would've been incorrect.

Anyway, no hard feelings since this is a genuine bug and you weren't too pushy about the AI PR. I generally appreciate issues over PRs these days, but this was also fine.

Enjoy using Mediabunny!! <3

Vanilagy (approved):

kzahel:
Great thank you so much. Yeah I had very little confidence in the root cause being correct but I think the value was in the test and also the fact that it was like a one-line diff is what gave me the courage to just send it over I'm glad I'm glad it annoyed you enough to make a proper fix

PR #318, opened 2026-03-07
MKV: use Cues for metadataOnly keyframe iteration

kzahel:
`getNextKeyPacket({ metadataOnly: true })` reads through every cluster even when the file has a Cues element. For MKV files served over HTTP, this means many unnecessary range requests just to get keyframe timestamps.

When `metadataOnly` is set and the track has cue points, this iterates cue points directly instead of reading clusters. Falls back to existing behavior when Cues aren't present.

Test included with a fixture that tracks I/O and confirms zero cluster reads during iteration.

### DIFF
--- a/test/node/matroska-metadata-seeking.test.ts
+++ b/test/node/matroska-metadata-seeking.test.ts
@@ -20,7 +20,9 @@ function createTrackingSource(buffer: Uint8Array) {
 		},
 	});
 
-	return { source, reads, clearReads: () => { reads.length = 0; } };
+	return { source, reads, clearReads: () => {
+		reads.length = 0;
+	} };
 }
 
 async function collectKeyframeTimestamps(


kzahel:
Hey sorry again for another drive by yeah again feel free to throw it away or whatever just treat this as a issue but I think the test fixture is valuable at least

Vanilagy:
I can't do this since the packet results must be guaranteed to be identical even when `metadataOnly: true` is set. The cues can be wrong! Or there can be multiple key frames in a cluster. But: Doing a cluster read feels like a non-issue since if you're gonna seek there, you'll need the video data anyway! So yeah, closing this one.

A potential way to solve this (maybe) is to, in the future, expose container-specific metadata; meaning, expose the cues directly.

kzahel:
Oh that makes sense okay I didn't realize the clusters and everything with MKV sorry about that yeah I think I think I'll it would be useful to just get the cue data and then I could sort of manually figure out byte ranges for streaming from that

Vanilagy:
Why do you need to do that

kzahel:
For knowing which segments (segment clusters) to download first to support streaming playback 

https://github.com/kzahel/JSTorrent/commit/ba1ae2fa10df1ee73ba61f2f448c6573194d1086

I think it should work fine and just download the whole cluster when seeking. But I haven't tested it e2e yet

Vanilagy:
Mediabunny automatically streams in the correct clusters when seeking, so why do you need manual logic for it? Just curious!

kzahel:
The reason we want direct MKV Cues access is that we are not always relying on the browser to seek the original file directly. In some cases we need to drive our own streaming pipeline for HLS/remux/transcode compatibility. For that, we need a cheap coarse seek index (timestamp -> cluster offset) so we can prioritize the right torrent pieces and build segment plans without scanning clusters across the whole file

Vanilagy:
Ah I see. Honestly doesn't sound like Mediabunny is of much help there, since it doesn't even provide you with the cluster offsets; it abstracts that away.

kzahel:
We do blocking reads against torrent data, so every read MediaBunny makes can trigger more downloading. For MKV, the exact keyframe path isn’t really tenable for us because it ends up pulling data across the whole file. If there were a coarse metadata API, like exposing Cues/cluster offsets directly, that would work perfectly and would let us delete our extra parser.

If it doesn't fit into your overall architecture and vision that's fine, I'll continue to parse the mvk cues etc myself.

Vanilagy:
Yah Mediabunny doesn't expose any container-specific data right now except for metadata tags

PR #324, opened 2026-03-13
Handle cross-realm ArrayBuffers in buffer source validation

ZerGo0:
TODO: Check if there are any other options to achieve this without having to change the entire project.

This PR fixes a quirk that currently fails the validations where cross-realm ArrayBuffers cause errors, but they are technically fine.

I added tests for this, but here is also a html version of it:
```
<!doctype html>
<meta charset="utf-8" />
<pre id="out"></pre>
<script type="module">
  const out = document.getElementById('out');
  const log = (value) => (out.textContent += `${value}\n`);

  const { validateVideoChunkMetadata } = await import(
    'https://unpkg.com/mediabunny@1.39.1/dist/modules/src/codec.js'
  );

  const iframe = document.createElement('iframe');
  iframe.srcdoc = '<!doctype html>';
  document.body.appendChild(iframe);
  await new Promise((resolve) => (iframe.onload = resolve));

  const crossRealmBuffer = new iframe.contentWindow.ArrayBuffer(4);

  log(`instanceof ArrayBuffer: ${crossRealmBuffer instanceof ArrayBuffer}`);
  log(`toString tag: ${Object.prototype.toString.call(crossRealmBuffer)}`);

  try {
    validateVideoChunkMetadata({
      decoderConfig: {
        codec: 'avc1.42001f',
        codedWidth: 2,
        codedHeight: 2,
        description: crossRealmBuffer,
      },
    });
    log('validation passed');
  } catch (error) {
    log(String(error));
  }

  iframe.remove();
</script>
```

Context: I work on browser extensions and I always have to "fight" with the browser sandbox when it comes to this. This issue only occurred in Firefox, Chrome and Safari pass these checks just fine. The HTML reprod fail in all browsers tho, I didn't have time to create a browser extension reprod, but I don't think it necessary.

### DIFF
--- a/src/adts/adts-demuxer.ts
+++ b/src/adts/adts-demuxer.ts
@@ -23,6 +23,7 @@ import {
 	AsyncMutex,
 	binarySearchExact,
 	binarySearchLessOrEqual,
+	isPromiseLike,
 	UNDETERMINED_LANGUAGE,
 } from '../misc';
 import { EncodedPacket, PLACEHOLDER_DATA } from '../packet';
@@ -84,7 +85,7 @@ export class AdtsDemuxer extends Demuxer {
 			// Skip all ID3v2 tags at the start of the file
 			while (true) {
 				let slice = this.reader.requestSlice(this.lastLoadedPos, ID3_V2_HEADER_SIZE);
-				if (slice instanceof Promise) slice = await slice;
+				if (isPromiseLike(slice)) slice = await slice;
 
 				if (!slice) {
 					this.lastSampleLoaded = true;
@@ -105,7 +106,7 @@ export class AdtsDemuxer extends Demuxer {
 			MIN_ADTS_FRAME_HEADER_SIZE,
 			MAX_ADTS_FRAME_HEADER_SIZE,
 		);
-		if (slice instanceof Promise) slice = await slice;
+		if (isPromiseLike(slice)) slice = await slice;
 		if (!slice) {
 			this.lastSampleLoaded = true;
 			return;
@@ -176,7 +177,7 @@ export class AdtsDemuxer extends Demuxer {
 
 			while (true) {
 				let headerSlice = this.reader.requestSlice(currentPos, ID3_V2_HEADER_SIZE);
-				if (headerSlice instanceof Promise) headerSlice = await headerSlice;
+				if (isPromiseLike(headerSlice)) headerSlice = await headerSlice;
 				if (!headerSlice) break;
 
 				const id3V2Header = readId3V2Header(headerSlice);
@@ -185,7 +186,7 @@ export class AdtsDemuxer extends Demuxer {
 				}
 
 				let contentSlice = this.reader.requestSlice(headerSlice.filePos, id3V2Header.size);
-				if (contentSlice instanceof Promise) contentSlice = await contentSlice;
+				if (isPromiseLike(contentSlice)) contentSlice = await contentSlice;
 				if (!contentSlice) break;
 
 				parseId3V2Tag(contentSlice, id3V2Header, this.metadataTags);
@@ -292,7 +293,7 @@ class AdtsAudioTrackBacking implements InputAudioTrackBacking {
 			data = PLACEHOLDER_DATA;
 		} else {
 			let slice = this.demuxer.reader.requestSlice(rawSample.dataStart, rawSample.dataSize);
-			if (slice instanceof Promise) slice = await slice;
+			if (isPromiseLike(slice)) slice = await slice;
 
 			if (!slice) {
 				return null; // Data didn't fit into the rest of the file
--- a/src/conversion.ts
+++ b/src/conversion.ts
@@ -40,6 +40,7 @@ import {
 	assert,
 	clamp,
 	isIso639Dash2LanguageCode,
+	isPromiseLike,
 	MaybePromise,
 	normalizeRotation,
 	promiseWithResolvers,
@@ -1292,7 +1293,7 @@ export class Conversion {
 			finalSamples = [sample];
 		} else {
 			let processed = trackOptions.process(sample);
-			if (processed instanceof Promise) processed = await processed;
+			if (isPromiseLike(processed)) processed = await processed;
 
 			if (!Array.isArray(processed)) {
 				processed = processed === null ? [] : [processed];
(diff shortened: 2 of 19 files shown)


Vanilagy:
Hey! Super niche issue but good that you caught this.

However, I feel for completeness sake, shouldn't we do this for *all* native classes? For example, this here also resolves to `false`:
```ts
new iframe.contentWindow.EncodedVideoChunk({
  data: new ArrayBuffer(0),
  timestamp: 0,
  type: 'key',
}) instanceof EncodedVideoChunk;
```

Meaning `EncodedPacket.fromEncodedChunk()` will fail for example.

Doing this just for `ArrayBuffer` feels incomplete, but doing it for *all* classes sounds like a slippery slope to me.

Vanilagy:
Spun up Claude to look for instances like these in the codebase:
<details>
src/id3.ts                                                                                                                                                                
                                                                                                                                                                            
  - L769: value instanceof Uint8Array                                                                                                                                       
                  
  src/packet.ts

  - L93: instanceof Uint8Array                                                                                                                                              
  - L114: instanceof Uint8Array
  - L222: instanceof EncodedVideoChunk, instanceof EncodedAudioChunk                                                                                                        
  - L258: instanceof Uint8Array                                                                                                                                             
                                                                                                                                                                            
  src/target.ts                                                                                                                                                             
                  
  - L108: instanceof WritableStream                                                                                                                                         
   
  src/metadata.ts                                                                                                                                                           
                  
  - L117, 144, 213, 236: instanceof Uint8Array                                                                                                                              
  - L199: instanceof Date
                                                                                                                                                                            
  src/misc.ts     
                                                                                                                                                                            
  - L166: x instanceof ArrayBuffer                                                                                                                                          
  - L167: x instanceof SharedArrayBuffer
                                                                                                                                                                            
  src/source.ts   

  - L111: buffer instanceof ArrayBuffer
  - L112: buffer instanceof SharedArrayBuffer
  - L176: blob instanceof Blob
  - L285: error instanceof Error
  - L296: src instanceof Request                                                                                                                                            
  - L389: url instanceof URL
  - L390: url instanceof Request                                                                                                                                            
  - L846: data instanceof Uint8Array
  - L860: data instanceof ReadableStream
  - L879: value instanceof Uint8Array                                                                                                                                       
  - L963: stream instanceof ReadableStream                                                                                                                                  
                                                                                                                                                                            
  src/sample.ts                                                                                                                                                             
                  
  - L61: value.data instanceof VideoFrame                                                                                                                                   
  - L74: value.data instanceof AudioData
  - L278: data instanceof ArrayBuffer                                                                                                                                       
  - L279: data instanceof SharedArrayBuffer                                                                                                                                 
  - L363: data instanceof VideoFrame
  - L399–404: instanceof HTMLImageElement, SVGImageElement, ImageBitmap, HTMLVideoElement, HTMLCanvasElement, OffscreenCanvas                                               
  - L499, 587, 632, 671, 991: this._data instanceof Uint8Array                                                                                                              
  - L774: context instanceof CanvasRenderingContext2D                                                                                                                       
  - L777: context instanceof OffscreenCanvasRenderingContext2D                                                                                                              
  - L866: context instanceof CanvasRenderingContext2D                                                                                                                       
  - L869: context instanceof OffscreenCanvasRenderingContext2D
  - L1104: x instanceof VideoFrame                                                                                                                                          
  - L1454: init.data instanceof ArrayBuffer                                                                                                                                 
  - L1742: this._data.buffer instanceof ArrayBuffer
  - L1788, 1834: audioBuffer instanceof AudioBuffer                                                                                                                         
  - L1951: x instanceof AudioData
                                                                                                                                                                            
  src/media-source.ts
                                                                                                                                                                            
  - L993: canvas instanceof HTMLCanvasElement
  - L994: canvas instanceof OffscreenCanvas
  - L1103, 2058: track instanceof MediaStreamTrack                                                                                                                          
  - L1276, 2155, 2296: error instanceof DOMException                                                                                                                        
  - L1997: audioBuffer instanceof AudioBuffer                                                                                                                               
                                                                                                                                                                            
  src/conversion.ts                                                                                                                                                         
                  
  - L1306: x instanceof VideoFrame                                                                                                                                          
   
  src/isobmff/isobmff-boxes.ts                                                                                                                                              
                  
  - L1472, 1617: value instanceof Uint8Array                                                                                                                                
   
  src/isobmff/isobmff-demuxer.ts                                                                                                                                            
                  
  - L2335, 2361, 2378: data instanceof Uint8Array                                                                                                                           
                  
  src/matroska/matroska-muxer.ts                                                                                                                                            
                  
  - L558: value instanceof Uint8Array

  src/matroska/ebml.ts                                                                                                                                                      
   
  - L403: data instanceof Uint8Array                                                                                                                                        
  - L447: data.data instanceof Uint8Array
  </details>

There's quite a few. Especially having support for the media elements sounds like a useful idea; e.g., creating a VideoSample from a video element in another iframe. Sounds crazy, but no reason that it shouldn't work.

This list does NOT exclude all of the `instanceof Promise` checks that are littered throughout the codebase, although I think it's safe to leave them as they are.

ZerGo0:
I'm all for a proper/full fix, but I feel like this would be pretty restricting in the future and hard to maintain, no?

ZerGo0:
Ok turns out that a lot of stuff needs to be changed. 93b636b0ccf61ce991659cab1718405b427ece06 contains a bunch of changes already, but I'm still not done, I'm still getting `Permission denied to access property "constructor"` somewhere.

So yea idk, if this even makes sense to support.

(I'm currently trying to extract audio (mp3) from a video (mp4), but I need to test a lot more stuff.)

Vanilagy:
I don't think we need `isPromiseLike`, that adds too much churn and the code paths are supposed to be fast. But I agree, it's kinda annoying that this has to be done, but how else would one do it? ArrayBuffers at least are trivially copyable, the same is not true for things like video elements.

ZerGo0:
I'm going to close this for now until I find a better solution.

ZerGo0:
For anyone having a similar issue, I ended up injecting a mediabunny script in the main world and communicating with it via `window.postMessage`. This is not ideal obviously, but works and is probably way easier than making mediabunny cross realm compatible.

Vanilagy:
The good ending.

ZerGo0:
Also, browser extension sandboxes are the worst sometimes... Gl everyone working with them

PR #325, opened 2026-03-17
FLAC: Add `appendOnly` option for FlaxMuxer, fix FlacDemuxer reading it

JonnyBurger:
Sample file generated: [sample-append-only.zip](https://github.com/user-attachments/files/26060876/sample-append-only.zip) (rename extension to .flac, since GitHub does not support .flac file extension)

- Implemented `appendOnly` option which writes `minimumFrameSize = 16`, `maximumFrameSize = 65535`, `minimumBlockSize = 0`, `maximumBlockSize = 0`. Spec allows setting `0` as unknown for block sizes, but not for frame sizes. But frame sizes are 16-bit (Spec: 8.2 Streaminfo)
- Manual test: Created a test file, put it in media-analyzer.pro and played it in Chrome.
- Added a test to flac.test.ts

This discovered that the demuxer does not handle `maximumBlockSize = 0`.
I fixed it as well, and added links to the spec to back up the reason for the values chosen.

### DIFF
--- a/docs/guide/output-formats.md
+++ b/docs/guide/output-formats.md
@@ -303,9 +303,16 @@ const output = new Output({
 The following options are available:
 ```ts
 type FlacOutputFormatOptions = {
+	appendOnly?: boolean;
+
 	onFrame?: (data: Uint8Array, position: number) => unknown;
 };
 ```
+- `appendOnly`\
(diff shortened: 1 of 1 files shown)


Vanilagy:
Very nice, thank you! Looks clean to me. Wasn't there an issue related to this feature? I can't for the life of me find it!!

Issue #329, opened 2026-03-18
Undesired video rotation when compressing metata-rotated videos.


sweidac:
I just discovered that an old issue https://github.com/Vanilagy/mediabunny/issues/151 (which you got fixed in 1.20-ish) appeared again in similar matter.

This bug was introduced in the release 1.35.0, not being present <= 1.34.5.

# Reproduce
1. Run the test in my for https://github.com/sweidac/mediabunny/blob/main/test/browser/conversion-dimensions.test.ts with `npx vitest conversion-dimensions`.
2.1. The "resizing metadata-rotation" leads to incorrect visual size" test shows the problematic outcome 
2.2. The "resizing natural rotation leads to correct visual size" test shows the expected outcome. Here, I prepared the video with ffmpeg so that the rotation is removed from metadata and transposed the actual pixels

# Side info

The problematic video wa shot on an iPhone X, so this is no esotheric usecase.
In the output of ffprobe below, the interesting part is the metadata of Stream #0 which says

> Display Matrix: rotation of -90.00 degrees

```
$ ffprobe test/public 1080p-portrait-rotated.mp4
ffprobe version 8.1 Copyright (c) 2007-2026 the FFmpeg developers
  built with Apple clang version 17.0.0 (clang-1700.6.4.2)
  configuration: --prefix=/opt/homebrew/Cellar/ffmpeg/8.1 --enable-shared --enable-pthreads --enable-version3 --cc=clang --host-cflags= --host-ldflags= --enable-ffplay --enable-gpl --enable-libsvtav1 --enable-libopus --enable-libx264 --enable-libmp3lame --enable-libdav1d --enable-libvpx --enable-libx265 --enable-openssl --enable-videotoolbox --enable-audiotoolbox --enable-neon
  libavutil      60. 26.100 / 60. 26.100
  libavcodec     62. 28.100 / 62. 28.100
  libavformat    62. 12.100 / 62. 12.100
  libavdevice    62.  3.100 / 62.  3.100
  libavfilter    11. 14.100 / 11. 14.100
  libswscale      9.  5.100 /  9.  5.100
  libswresample   6.  3.100 /  6.  3.100
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '1080p-portrait-rotated.mp4':
  Metadata:
    major_brand     : mp42
    minor_version   : 0
    compatible_brands: isommp42
    creation_time   : 2025-09-26T09:57:19.000000Z
[...]

PR #330, opened 2026-03-19
Count flipX and flipY related data in transform matrix when calculating the rotation

tigran-balayan-rf:
Fix for the issue #329 
1. Rotation calculation fix - now counts also flip values from transform matrix
2. Conversion output metadata fix - now puts correct rotation value. Previously the value update was inside async function, which in a sync part of the code was causing to use the old value when putting into metadata object.


Vanilagy:
Fixed in https://github.com/Vanilagy/mediabunny/commit/0d50526ff65399e8c2e83c3fca43198eb9a00243

tigran-balayan-rf:
@Vanilagy I see that the PR covers the second fix, but not the 1st, are you sure it's not necessary, since the 1st fix is what I noticed first.
https://drive.google.com/file/d/1uObCr_y1nklvDT9FvSft8I_OfvZzuGMa/view?usp=sharing
compare this video in MediaBunny's media-player example and in regular video tag, you'll see the difference

Vanilagy:
@tigran-balayan-rf The matrix rotation fix seemed unrelated.

I did check your video, and you're right that it displays differently, but this is because the video actually has a flip to it. No amount of rotation is gonna add that to the video, so that means Mediabunny needs to be extended by a flip system. I'll look into it. Can you take a video of some text so I can verify this hypothesis?

tigran-balayan-rf:
@Vanilagy I'll try to record one tomorrow. And yes you're right, my solution doesn't fix it but fakes it and that effect can't be handled by rotation in anyway.

Vanilagy:
Dope :)

Vanilagy:
Could you still try to record a vid?

PR #341, opened 2026-04-01
Fix packets outside edit list range not being marked for discard

raedle:
## Problem

When an MP4 file has an edit list (`elst` box), some samples may have a
presentation timestamp that falls outside the edit list's valid range
`[mediaTime, mediaTime + segmentDuration)`. These samples should be
decoded (they may be referenced by other frames for inter-prediction)
but their decoded output should not be presented.

FFmpeg handles this by setting `AVINDEX_DISCARD_FRAME` on such samples
during edit list processing, which later becomes `AV_PKT_FLAG_DISCARD`
on the packet:
https://github.com/FFmpeg/FFmpeg/blob/d244d438c/libavformat/mov.c#L4417-L4418

Mediabunny had no equivalent logic — it used the edit list only to shift
timestamps but never checked whether a sample fell outside the valid
range. This caused two issues:
1. One extra presentable frame compared to FFmpeg (e.g. 229 vs 228)
2. If the discarded packet was skipped entirely (as in an earlier
   iteration of this fix), frames referencing it as a prediction
   reference would decode with artifacts

## Changes

- Store the edit list's `segmentDuration` from the `elst` box on
  `InternalTrack` and convert it to media timescale
- Add a `discard` boolean to `EncodedPacket` (defaults to `false`)
- In both `fetchPacketForSampleIndex` and `fetchPacketInFragment`,
  flag packets whose PTS falls outside the edit list range with
  `discard: true`
- Add a test fixture (`test/public/edit-list-discard.mp4`) — a
  synthetic video where the last sample's PTS lands exactly at the
  edit list end boundary

## Test plan

- [ ] Existing tests pass (`npx vitest --run`)
- [ ] Demux `test/public/edit-list-discard.mp4` and confirm 50 packets
      emitted with exactly 1 marked `discard: true` (sample 49)
- [ ] Demux a file without an edit list and confirm all packets have
      `discard: false`
- [ ] Decode a file with discarded packets end-to-end and verify no
      artifacts on surrounding frames


raedle:
Hey @Vanilagy! I ran into an issue with videos where packets had the discard flag set. This PR is an attempt to replicate the same logic ffmpeg uses to handle DISCARD packets. Not married to this particular solution, but mainly wanted to flag this and see if there are better approaches we should consider.

Vanilagy:
Hey, edit lists are indeed not fully supported in Mediabunny (as in, all the crazy things you can model with them aren't fully supported), but the normal use case is supported: Trim away the start of the media. This is described in the docs; if the timestamps land below zero, you are expected not to present the samples.

The file you attached in this PR is actually not a demonstration of this: the file you attached uses the old-fashioned non-negative composition time offsets, meaning the first sample actually starts at timestamp 1024 (in its timescale). To offset this back to zero, the edit list trims 1024 units off the start of the media. If you check your file with ffprobe, you'll see that none of the packets are discarded.

You should check if your files land in the "Beginning packets have negative timestamps" category; in this case, you're done: simply don't present those frames. If you have something more complicated, please send over the file!

Vanilagy:
Any updates?

Vanilagy:
Closing due to inactivity

PR #342, opened 2026-04-01
Fix live seek hangs in CanvasSink random access

sor4chi:
## Summary
Random seeks on live HLS through `CanvasSink.getCanvas()` could hang indefinitely because packet retrieval kept waiting for the live edge to advance.

I think we needs to pass `CanvasSink(videoTrack, { skipLiveWait: true })` for live HLS scrubbing, but the random-access path in `CanvasSink` / `VideoSampleSink` did not propagate `skipLiveWait` down to `getPacket`, `getKeyPacket`, and `getNextPacket`, so the option had no effect.

This change threads `skipLiveWait` through the random-access sink path so live seeks can resolve using the currently known packets only.

## Changes
- add `skipLiveWait` to `PacketRetrievalOptions`
- propagate `skipLiveWait` through `BaseMediaSampleSink` packet retrieval paths
- add `skipLiveWait` to `CanvasSinkOptions`
- forward `skipLiveWait` from `CanvasSink` to `VideoSampleSink`

### DIFF
--- a/src/media-sink.ts
+++ b/src/media-sink.ts
@@ -65,6 +65,13 @@ export type PacketRetrievalOptions = {
 	 * its data, this option cannot be enabled together with `metadataOnly`.
 	 */
 	verifyKeyPackets?: boolean;
+
+	/**
+	 * When querying packets in live media that are in the future relative to the current live edge, Mediabunny will,
+	 * by default, wait for the stream to advance until the query can be satisfied. Set this to `true` to only use the
+	 * currently known packets and never wait for the live edge to catch up.
+	 */
+	skipLiveWait?: boolean;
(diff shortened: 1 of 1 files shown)


sor4chi:
I'm sorry, I did not realize you were already working on it. I am looking forward to the upcoming HLS support.

Vanilagy:
No worries. `skipLiveWait` is already on `PacketRetrievalOptions`, but you actually raise a good point that there's no way to set it using the higher-level sinks. I'll fix that!

Vanilagy:
Fixed in the linked commit

sor4chi:
Thanks! I really appreciate it

PR #347, opened 2026-04-08
feat: expose buffered source and track ranges

### LINKED ISSUE #346: Expose buffered ranges for on-demand media
mediabunny already has:
- `Source.onread`
- internal source cache state
- enough ISOBMFF metadata to map byte offsets to media timestamps

But there is no public way to ask:
- which byte ranges are currently retained in the source cache
- which media time ranges are currently buffered for a track

`onread` is not enough for this, because it reports read history, not the ranges that are still retained after cache eviction.

## Proposal

Add two small public APIs:

```ts
type BufferedByteRange = { start: number; end: number };
type BufferedTimeRange = { start: number; end: number };
```

- `BufferedByteRange.start/end`: byte offsets
- `BufferedTimeRange.start/end`: media time in seconds

```ts
Source.getBufferedByteRanges(): BufferedByteRange[]
InputTrack.getBufferedTimeRanges(): Promise<BufferedTimeRange[]>
```

Expected semantics:
- end-exclusive
- sorted by start
- snapshot of the current retained cache state
- `getBufferedTimeRanges()` returns `[]` when unsupported

## Background

Browser media APIs already expose buffered media ranges:
- `HTMLMediaElement.buffered`
- `TimeRanges`

This makes it possible for applications to build buffered UI without depending on private internals.

mediabunny already tracks the necessary source cache state internally, so exposing buffered ranges would let applications build similar UI on top of the existing cache model.

References:
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/buffered
- https://developer.mozilla.org/en-US/docs/Web/API/TimeRanges
- https://www.w3.org/TR/media-source-2/

sor4chi:
Closes #346.

## Changes

- add `Source.getBufferedByteRanges()`
- add `InputTrack.getBufferedTimeRanges()`
- implement `InputTrack.getBufferedTimeRanges()` for on-demand ISOBMFF/MP4

## Tests

- `test/node/source-buffered-ranges.test.ts`
- `test/node/buffered-time-ranges.test.ts`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -127,6 +127,7 @@ export {
 } from './output';
 export {
 	Source,
+	BufferedByteRange,
 	BlobSource,
 	BlobSourceOptions,
 	BufferSource,
@@ -173,6 +174,7 @@ export {
 	InputTrack,
 	InputVideoTrack,
 	InputAudioTrack,
+	BufferedTimeRange,
 	PacketStats,
 } from './input-track';
 export {
--- a/src/input-track.ts
+++ b/src/input-track.ts
@@ -30,6 +30,16 @@ export type PacketStats = {
 	averageBitrate: number;
 };
 
+/**
+ * A buffered media time range in seconds. The end timestamp is exclusive.
+ * @group Input files & tracks
+ * @public
+ */
+export type BufferedTimeRange = {
+	start: number;
+	end: number;
+};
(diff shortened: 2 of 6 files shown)


Vanilagy:
See comment on linked issue

PR #349, opened 2026-04-16
Add BufferTarget.onFinalize, OutputOptions.onFinalize, ConcurrentRunner

ahmedrowaihi:
The `finalized` event is sync — the muxer doesn't await listeners. So if I do an async upload inside it, it's fire-and-forget. Segments pile up in memory faster than they upload, and long videos OOM.

`StreamTarget` solves this for servers that accept streaming bodies, but not for S3 `PutObject` (needs `Content-Length`, SDK wants a `Blob` body). That means `BufferTarget` is the only option, and there's no way to plug an awaitable upload into it.

This PR adds an `onFinalize` option to `BufferTarget` that the muxer awaits:

```ts
new BufferTarget({
  onFinalize: async (buffer) => {
    await fetch('/upload', { method: 'PUT', body: buffer });
  },
});
```

With `PathedTarget`, the next segment won't start until the current one is uploaded. Memory stays flat.

Shape matches the `HlsOutputFormatOptions` callbacks (`onSegment`, `onMaster`, etc). Constructor still accepts no args, so no breaking change. Only touching `BufferTarget` — `StreamTarget` and `FilePathTarget` already have async I/O paths.

Using this in production — before: OOM on 3-minute 1080p. After: flat memory regardless of length. Happy to change the shape if you'd prefer something different.

### DIFF
--- a/docs/guide/writing-hls.md
+++ b/docs/guide/writing-hls.md
@@ -195,14 +195,7 @@ await output.finalize();
 // All files have been uploaded to the server
 ```
 
-[`runner.parallelism`](../api/ConcurrentRunner#parallelism) is mutable if the optimal concurrency changes over time — for example, to back off on server errors or probe higher when latency is stable:
-
-```ts
-runner.parallelism = 1; // Back off after a 5xx or 429
-runner.parallelism = 8; // Probe up when latency is low
-```
-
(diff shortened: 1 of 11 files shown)


ahmedrowaihi (commented):
Image

Vanilagy:
image

PR #352, opened 2026-04-20
Work around Chromium 147+ Canvas2D drawImage(VideoFrame) regression that produces black frames

alec-watts:
## Problem

Chromium 147 silently broke `Canvas2D.drawImage(VideoFrame)` (and `createImageBitmap(VideoFrame)`) for some hardware-decoded frames — most notably **HEVC Range Extensions** (`yuv422p10le` / `format === null`), but other 10-bit / 4:2:2 hardware-decoded frames are affected as well. The frame decodes fine, but every callsite that draws it through Canvas2D gets an **all-black** canvas. WebGL `texImage2D(VideoFrame)` is not affected — it uses a different upload path inside Blink.

In mediabunny this currently breaks `Conversion.init()` for affected sources on affected Chromium builds: every encoded output frame is black. The repro is essentially `Conversion.init({ input: hevc422p10leMov, output: mp4 })` — the file plays fine in `<video>`, decodes fine via `VideoDecoder`, but the resize/encode canvas comes out black.

## Bisect

I bisected Chrome for Testing builds and pinned the regression to a one-build window:

| Chrome build | `drawImage(VideoFrame)` for HEVC 4:2:2 10-bit |
| --- | --- |
| `147.0.7705.0` | ✅ pixels OK |
| `147.0.7706.0` | ❌ all-black |

The change of behavior in that window matches a Blink refactor of how `StaticBitmapImage`s are created from `VideoFrame`s on the Canvas2D path; WebGL was untouched. I'm filing this as a Chromium bug separately, but until/unless that ships a fix this is a working library-side workaround that costs nothing on non-affected configurations.

## Approach

The fix is intentionally surgical — a single new module plus a 4-line change to `VideoSample.toCanvasImageSource()`:

1. **Lazy per-format probe.** On the first `VideoFrame` of each distinct `VideoFrame.format` value (including `null`), draw it into a small throwaway Canvas2D and read a few pixels back. If every sampled pixel has zero RGB, mark that format as broken on this Chromium build.
2. **WebGL fallback for broken formats.** Re-render the `VideoFrame` through a shared WebGL helper (`texImage2D` → `drawArrays` into a reused `OffscreenCanvas` / `HTMLCanvasElement`) and return that canvas instead of the `VideoFrame` from `toCanvasImageSource()`.
3. **Single funnel.** Because every internal `drawImage` callsite (`drawWithFit`, `draw`, `CanvasSink`, the encoder resize path in `media-source.ts`) and any user-land code goes through `VideoSample.toCanvasImageSource()`, no per-callsite changes are needed.
4. **Auto-disables itself** if WebGL is unavailable or the shader/texture upload fails — the call falls through to the original (broken) Canvas2D path so non-affected sources keep working unchanged.

## Costs

| Browser / config | Overhead |
| --- | --- |
| Non-Chromium (Firefox, Safari, …) | None — early return on `!isChromium()` |
| Chromium, format not affected | One tiny `drawImage` + ≤4×4 `getImageData` per distinct `VideoFrame.format`, then nothing |
| Chromium, format affected | One shared `WebGLRenderingContext` + canvas + program, reused per-frame |

No public API changes. No new exports. No behavior change on browsers that aren't broken.

## Files

- `src/chromium-video-frame-canvas-fallback.ts` — new internal helper module (probe + WebGL render).
- `src/sample.ts` — `toCanvasImageSource()` now routes returned `VideoFrame`s through `maybeRewriteVideoFrameForCanvas2d()`.

## Verification

- `npx tsc --noEmit -p src/tsconfig.json` — clean.
- `npx eslint src/chromium-video-frame-canvas-fallback.ts src/sample.ts` — clean.
- Manual repro on Chrome 147.0.7707.x with an HEVC Rext `yuv422p10le` source (`C4340.MP4` from a Sony FX3): without the patch, the output proxy is black; with the patch, the output proxy contains correct pixels and audio. Same source on Chrome 147.0.7705.0 and on Firefox confirms the workaround correctly stays out of the way (probe returns "format ok") on those builds.

Happy to add a Vitest browser test that asserts `toCanvasImageSource()` round-trips a non-black bitmap through `getImageData` for a known-bad fixture if you'd like — let me know which fixture format you'd want me to commit.

### DIFF
--- a/src/chromium-video-frame-canvas-fallback.ts
+++ b/src/chromium-video-frame-canvas-fallback.ts
@@ -1,0 +1,236 @@
+/*!
+ * Copyright (c) 2026-present, Vanilagy and contributors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+import { isChromium } from './misc';
+
+/**
+ * Workaround for a Chromium 147 regression in which Canvas2D `drawImage(VideoFrame)`
+ * (and `createImageBitmap(VideoFrame)`) silently produces an all-black canvas for some
+ * hardware-decoded `VideoFrame`s — most notably HEVC Range Extensions (e.g. 4:2:2 10-bit,
+ * `format === null`), but other 10-bit / 4:2:2 hardware frames are also affected.
+ *
+ * Bisect of Chrome for Testing builds: 147.0.7705.0 PASSes, 147.0.7706.0 FAILs.
+ * The change of behavior corresponds to a Blink refactor of how `StaticBitmapImage`s
+ * are created from `VideoFrame`s (Canvas2D path). WebGL `texImage2D(VideoFrame)` still
+ * works correctly because it goes through a different upload path.
+ *
+ * Strategy:
+ *  1. On the first `toCanvasImageSource()` call for each distinct `VideoFrame.format`
+ *     value, we probe by drawing the frame into a tiny throwaway Canvas2D and reading
+ *     a few pixels back. If every sampled pixel has zero RGB, we mark this format as
+ *     "broken on this Chromium build".
+ *  2. For broken formats, we reroute via a shared WebGL helper that uploads the
+ *     `VideoFrame` to a texture and renders it to an `OffscreenCanvas` /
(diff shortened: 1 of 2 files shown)


Vanilagy:
Closing this for now due to https://issues.chromium.org/issues/504680933

Should be fixed in Chrome 148 again. Not worth it for me "polluting" Mediabunny's codebase with a temporary workaround until then. Soon, hopefully, this bug will have been forgotten!

Can reopen if urgent

PR #353, opened 2026-04-20
Include actual HTTP status code when unexpectedly did not get 206

JonnyBurger:
Bug https://discord.com/channels/809501355504959528/1445495265028079718/1445556450234466345 seems to affect a couple of people.

For now I just would like to surface a bit more information to learn more about the problem if it happens again in the future.


Vanilagy:
This code is no longer there in the HLS branch (which will soon be merged into main), so no need for this fortunately!

PR #359, opened 2026-04-26
Add sidx box parsing and expose via Input.getSegmentIndex

### LINKED ISSUE #358: Add `sidx` parsing to the ISOBMFF demuxer
# Add `sidx` parsing to the ISOBMFF demuxer

The muxer writes `sidx` ([`isobmff-boxes.ts`](https://github.com/Vanilagy/mediabunny/blob/main/src/isobmff/isobmff-boxes.ts)) but the demuxer doesn't read it back. There's no way today to get per-subsegment byte offsets / durations / SAP markers from a fragmented MP4.

Worth filling because:

- It's the structure DASH on-demand profile MPDs and HLS `#EXT-X-BYTERANGE` references derive from. Anyone generating those manifests from an existing fMP4 needs the entries.
- HLS muxer in [#291](https://github.com/Vanilagy/mediabunny/pull/291) already writes single-file byterange HLS. Reading the same shape post-hoc has no equivalent path because `sidx` isn't surfaced — the only way mediabunny gets segment ranges out of an fMP4 today is by parsing them out of an existing m3u8.
- Once exposed, `seek` / `computeDuration` / `computePacketStats` could use it as a fast index when present. Out of scope here — this proposal is only the parser + accessor. Wiring it into existing read paths would be follow-up work.

## Proposed API

Mirroring the `pssh` flow:

```ts
// src/isobmff/isobmff-misc.ts
export type SidxReference = {
  referenceType: 0 | 1;
  referencedSize: number;
  subsegmentDuration: number;
  startsWithSAP: 0 | 1;
  sapType: number;
  sapDeltaTime: number;
};

export type SidxBox = {
  referenceID: number;
  timescale: number;
  earliestPresentationTime: number;
  firstOffset: number;
  references: SidxReference[];
  boxStart: number;
  boxSize: number;
};

export const parseSidxBoxContents = (contents: Uint8Array, version: number): SidxBox;
```

```ts
// On Input
async getSegmentIndex(): Promise<SidxBox[]>;
```

Multiple `sidx` boxes are valid (one per `reference_ID`); caller maps to tracks or a per-track helper is added.

## Cost

No regressions. Files without `sidx`: zero impact (case never fires). Files with `sidx`: one extra parse on open, ~7 KB / 600 refs for a 1 h CMAF, microseconds. Today the box is silently skipped.

Spec: ISO/IEC 14496-12 §8.16.3. Box switch lives next to `'pssh'`, `'pasp'`, `'tenc'`. Happy to send a PR — let me know whether you'd want it against `main` or rebased onto the `hls` branch (#291), since the demuxer is being heavily modified there.

ahmedrowaihi:
Closes #358.

Mirrors the existing `pssh` flow:

- `SidxBox` + `SidxReference` types and `parseSidxBoxContents` in `isobmff-misc.ts`.
- Inline branch in `IsobmffDemuxer.readMetadata` parses any `sidx` encountered during the file-root walk; stores onto a new `sidxBoxes` field. The walk now continues past `moov` so a `sidx` placed after it (DASH on-demand profile layout) is captured too. The `moof` branch breaks early when `movieTimescale !== -1` so a self-contained fragmented file doesn't trip the "no init file" error after parsing its own moov.
- `Demuxer.getSegmentIndex()` defaults to `[]`; ISOBMFF overrides. Other demuxers inherit the default.
- `Input.getSegmentIndex()` delegates to the demuxer.
- Roundtrip + non-fragmented tests in `test/browser/cmaf.test.ts`.

No behaviour changes for existing callers — `sidx` was previously silently skipped, now stored.

I made a little visualizer, so it can show how this box helps
image

Branched off `hls` since the demuxer is being heavily modified there. Happy to rebase onto `main` if you'd prefer.

### DIFF
--- a/src/isobmff/isobmff-demuxer.ts
+++ b/src/isobmff/isobmff-demuxer.ts
@@ -64,7 +64,14 @@ import {
 	HEX_STRING_REGEX,
 } from '../misc';
 import { EncodedPacket, PLACEHOLDER_DATA } from '../packet';
-import { buildIsobmffMimeType, parsePsshBoxContents, psshBoxesAreEqual, PsshBox } from './isobmff-misc';
+import {
+	buildIsobmffMimeType,
+	parsePsshBoxContents,
+	parseSidxBoxContents,
+	psshBoxesAreEqual,
+	PsshBox,
+	SidxBox,
+} from './isobmff-misc';
 import {
 	MAX_BOX_HEADER_SIZE,
 	MIN_BOX_HEADER_SIZE,
@@ -303,6 +310,7 @@ export class IsobmffDemuxer extends Demuxer {
 	isFragmented = false;
 	fragmentTrackDefaults: FragmentTrackDefaults[] = [];
 	psshBoxes: PsshBox[] = [];
+	sidxBoxes: SidxBox[] = [];
 	currentFragment: Fragment | null = null;
 	/**
 	 * Caches the last fragment that was read. Based on the assumption that there will be multiple reads to the
@@ -344,6 +352,11 @@ export class IsobmffDemuxer extends Demuxer {
 		return this.metadataTags;
 	}
 
+	override async getSegmentIndex() {
+		await this.readMetadata();
+		return this.sidxBoxes;
+	}
+
 	readMetadata() {
 		return this.metadataPromise ??= (async () => {
 			let currentPos = 0;
@@ -363,6 +376,16 @@ export class IsobmffDemuxer extends Demuxer {
 				if (boxInfo.name === 'ftyp' || boxInfo.name === 'styp') {
 					const majorBrand = readAscii(slice, 4);
 					this.isQuickTime = majorBrand === 'qt  ';
+				} else if (boxInfo.name === 'sidx') {
+					let sidxSlice = this.reader.requestSlice(slice.filePos, boxInfo.contentSize);
+					if (sidxSlice instanceof Promise) sidxSlice = await sidxSlice;
+					if (!sidxSlice) break;
+
+					this.sidxBoxes.push(parseSidxBoxContents(
+						readBytes(sidxSlice, boxInfo.contentSize),
+						startPos,
+						boxInfo.totalSize,
+					));
 				} else if (boxInfo.name === 'moov') {
 					// Found moov, load it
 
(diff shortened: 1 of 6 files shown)


ahmedrowaihi:
hey @vitalygashkov I saw ur [dasha](https://github.com/azot-labs/dasha) I feel like you would love this !

src/isobmff/isobmff-misc.ts
```diff
@@ -88,3 +88,117 @@ export const psshBoxesAreEqual = (a: PsshBox, b: PsshBox) => (
...
+		pos += 8;
+	}
+
+	pos += 2; // Reserved
+	const referenceCount = view.getUint16(pos);
+	pos += 2;
+
+	const references: SidxReference[] = [];
+	for (let i = 0; i < referenceCount; i++) {
+		const sizeWord = view.getUint32(pos);
```
ali-sao, line 186: Guard `referenceCount` against available buffer. The function should either bound-check (referenceCount * 12 <= view.byteLength - pos) and throw a descriptive error, or stop iterating with a warning in case of a malfromed or truncated file. Assume `pos + 12 > view.byteLength`

src/isobmff/isobmff-demuxer.ts
```diff
@@ -385,8 +408,16 @@ export class IsobmffDemuxer extends Demuxer {
...
 
-					break;
+					// Don't break here — keep iterating so a `sidx` that follows `moov`
+					// (DASH on-demand profile layout) is also captured. Iteration stops
+					// when we hit `moof`/`mdat` or EOF.
 				} else if (boxInfo.name === 'moof') {
+					if (this.movieTimescale !== -1) {
+						// `moov` was already parsed earlier in the same file; the rest is
+						// fragments which are read lazily.
+						break;
```
ali-sao, line 418: I'd recommend explicitly testing the self-contained fragmented MP4 case (one file with ftyp + moov + sidx + moof + mdat)

test/browser/cmaf.test.ts
```diff
@@ -85,6 +85,97 @@ test('CMAF with video track', async () => {
...
+		formats: ALL_FORMATS,
+	});
+	using segmentInput = new Input({
+		source: new BufferSource(output.target.buffer!),
+		formats: ALL_FORMATS,
+		initInput,
+	});
+
+	const sidxBoxes = await segmentInput.getSegmentIndex();
+	expect(sidxBoxes.length).toBeGreaterThan(0);
```
ali-sao, line 141: This alongside with the rest of خpositive assertions are a bit weak. 
Thiking. out loud, we should be more precise about what we parsed, not if we are able to parse, tests to cover the following scenarios are also required :

- The sum of `referencedSize` values matches the `mdat` size (or the segment minus init).
- `referenceID` matches the video track ID known from the init segment.
- `timescale` matches the track's `mdhd` timescale.
- `earliestPresentationTime` is 0 (or a known value) for the first segment.

src/isobmff/isobmff-misc.ts
ahmedrowaihi: my director is here 😎 what a surprise

@Vanilagy what's ur thought ? do u prefer warning skip or a throw ?

src/isobmff/isobmff-demuxer.ts
ahmedrowaihi: done

test/browser/cmaf.test.ts
ahmedrowaihi: yes this is more accurate, implemented

ali-sao:
Excellent job @ahmedrowaihi

Vanilagy:
See https://github.com/Vanilagy/mediabunny/issues/358#issuecomment-4328720572

In the future, let's discuss this in the issue before you create a PR; less work for both of us!

Issue #371, opened 2026-05-11
`edts` leads to non-seamless audio decoding


JonnyBurger:
We are interested in correctly lining up audio from multiple clips to prevent little crackles.

I've prepared 2 audio sources, both MP4+AAC:

1. Conventional sine wave with edts offset of 1024:
```tsx
const URL = 'https://remotion.media/video.mp4'
```

2. A bit more unconventional, edts offset of 2112:
```tsx
const URL = 'https://remotion.media/edts.mp4';
```

(1) is a reference and does not have the bug I am describing for (2).

Extracting samples from 0-4sec, then from 4-8sec leads to:

```tsx
import {ALL_FORMATS, AudioBufferSink, Input, UrlSource} from 'mediabunny';
import {test} from 'vitest';

test('edts', async () => {
	const source = new UrlSource(URL);
	const input = new Input({
		source,
		formats: ALL_FORMATS,
	});
	const audioTrack = await input.getPrimaryAudioTrack();
	if (!audioTrack) {
		throw new Error('No audio track found');
	}

	const audioBufferSink = new AudioBufferSink(audioTrack);
	const buffers = audioBufferSink.buffers(0, 4);
	for await (const buffer of buffers) {
		console.log(buffer.timestamp);
	}

	console.log('----');

	const buffers2 = audioBufferSink.buffers(4, 8);
	for await (const buffer of buffers2) {
		console.log(buffer.timestamp);
	}
});
```

```
3.947392290249433
3.9706122448979593
3.9938321995464854 // do not line up
----
3.9923809523809526 // do not line up
4.015600907029478
4.038820861678005
```

Bug does not happen for `video.mp4`
```
3.9466666666666668
3.968
3.989333333333333 // do line up
----
3.989333333333333 // do line up
4.010666666666666
4.032
```

I can understand imprecision for Matroska files, but from what I understand, it should be possible to perfectly line up MP4 files.

Vanilagy:
Haha, this is also now fixed due to https://github.com/Vanilagy/mediabunny/issues/370#issuecomment-4443396704:

<img width="351" height="296" alt="Image" src="https://github.com/user-attachments/assets/be03ec7f-6e0f-4cc2-a3c6-399c38b61abd" />

Thanks for the easy to copy repro!!

Note that you will probably need to start the second decode a tad earlier due to the AAC decoder warmup; you know the drill. I'd start it 1024 samples ealier, then the two waves should line up continuously. But, would need to test that to confirm.

PR #372, opened 2026-05-11
Fix MP4 AAC edit-list seek seams

JonnyBurger:
Fixes bug described in #371.

Made using coding agent - reviewed manually, and it seems sensible.


Vanilagy:
https://github.com/Vanilagy/mediabunny/issues/371#issuecomment-4443474641

PR #373, opened 2026-05-11
Fix MP4 AAC edit-list audio start

### LINKED ISSUE #370: `edts` is probably not correctly respected during audio decoding
We are interested in correctly lining up audio from multiple clips to prevent little crackles.

I've prepared 2 audio sources, both MP4+AAC:

1. Conventional sine wave with edts offset of 1024:
```tsx
const URL = 'https://remotion.media/video.mp4'
```

2. A bit more unconventional, edts offset of 2112:
```tsx
const URL = 'https://remotion.media/edts.mp4';
```

(1) renders the waveform as expected.

With (2), we seem to deviate from other video editing programs.

In Davinci Resolve, the waveform starts at 0.

Image

In Diffusion Studio and Remotion, both powered by Mediabunny, we have the same behaviour - not starting at 0.

Image
Image

A test confirms this:

```tsx
import {ALL_FORMATS, AudioBufferSink, Input, UrlSource} from 'mediabunny';
import {test} from 'vitest';

test('edts', async () => {
	const source = new UrlSource(URL);
	const input = new Input({
		source,
		formats: ALL_FORMATS,
	});
	const audioTrack = await input.getPrimaryAudioTrack();
	if (!audioTrack) {
		throw new Error('No audio track found');
	}

	const audioBufferSink = new AudioBufferSink(audioTrack);
	const buffers = audioBufferSink.buffers(0, 1);
	const next = await buffers.next();
	console.log(next.value!.buffer.getChannelData(0).slice(0, 50));
	// [0, 0, 0, 0, 0, 0, 0] for edts.mp4
    // [-0.000007795611054461915,-0.000024478687919327058] for video.mp4
});
```

I suspect an edts offset is hardcoded or we in some other way do not handle it correctly.

I will run an agent to try to try to solve this issue.

JonnyBurger:
## Summary
- Decode through negative AAC edit-list priming when a requested range starts inside it.
- Trim decoded audio samples that cross timestamp 0 so edited MP4 audio begins with presented waveform data.

Fixes #370.

## Stack note
This is logically stacked on #372, but GitHub cannot use the fork branch from #372 as the base branch in Vanilagy/mediabunny. Please review/merge #372 first; this PR currently shows both commits when compared against main.

## Test plan
- [x] npx tsc --noEmit --project tsconfig.json
- [x] npx eslint src/media-sink.ts test/browser/mp4-audio.test.ts
- [x] npx vitest --run --project browser test/browser/mp4-audio.test.ts


JonnyBurger:
This is more than I wanted to change.
Could not test this and think about it enough that I wanna submit it as PR right now.

But maybe useful as this makes the tests pass while `main` does not.

PR #379, opened 2026-05-14
fix: add integer validation to isU32 utility function

ptanh05:
## Summary
- The `isU32()` function in `src/misc.ts` was missing integer validation, allowing floating-point numbers to pass as valid unsigned 32-bit integers
- Added `Number.isInteger()` check to ensure only actual integers are accepted
- Added unit tests covering edge cases: valid u32 values, negative numbers, boundary values, and floating-point rejection

## Test plan
- [x] Added `test/node/isU32.test.ts` with comprehensive test coverage
- [x] All existing tests pass
- [x] New tests confirm the fix handles floating-point inputs correctly

## Impact
This bug could cause subtle issues in binary parsing contexts where `isU32()` is used for input validation. Floating-point values would incorrectly pass validation, potentially leading to data corruption or unexpected behavior.

### DIFF
--- a/src/misc.ts
+++ b/src/misc.ts
@@ -38,7 +38,7 @@ export const last = <T>(arr: T[]) => {
 };
 
 export const isU32 = (value: number) => {
-	return value >= 0 && value < 2 ** 32;
+	return Number.isInteger(value) && value >= 0 && value < 2 ** 32;
 };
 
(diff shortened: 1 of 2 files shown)


Vanilagy:
It's an internal function, doesn't need validation.

PR #380, opened 2026-05-14
Fix ReDoS vulnerability in PCM codec regex

ptanh05:
## Summary

- Fix catastrophic backtracking (ReDoS) in `PCM_CODEC_REGEX` caused by redundant outer quantifier on capturing group: `/^pcm-([usf])(\d+)+(be)?$/`
- The pattern `(\d+)+` creates exponential backtracking on crafted input like `"pcm-u999...999x"`
- Fixed by removing the unnecessary outer `+`, changing to `(\d+)`
- Add regression test with crafted input to ensure linear matching time

### DIFF
--- a/src/codec.ts
+++ b/src/codec.ts
@@ -629,7 +629,7 @@ export const guessDescriptionForAudio = (decoderConfig: AudioDecoderConfig): Uin
 
 export const OPUS_SAMPLE_RATE = 48_000;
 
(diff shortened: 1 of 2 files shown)


Vanilagy:
A ReDoS this certainly this not, there are no "crafted inputs" reaching that regex, but the regex was indeed redundant, thanks

PR #385, opened 2026-05-22
Fix MediaStreamTrackProcessor worker support handshake

jfrolich:
## Summary
- Fixes a race in the `MediaStreamTrackProcessor` worker-support probe used by `MediaStreamVideoTrackSource`.
- Changes the helper worker handshake from an immediate startup message to an explicit request/response after the main thread listener is registered.
- Prevents `mediaStreamTrackProcessorIsSupportedInWorker()` from hanging if the worker posts its support result before the caller starts listening.

## Bug
When `MediaStreamTrackProcessor` is unavailable on the main thread, `MediaStreamVideoTrackSource._start()` checks whether it is available inside a helper worker. The worker currently posts its `{ type: 'support' }` response immediately when it starts.

The main thread creates the worker first and only then attaches the `message` listener inside `mediaStreamTrackProcessorIsSupportedInWorker()`. If the worker starts quickly enough, the support message can be sent before that listener exists. In that case the promise returned by `mediaStreamTrackProcessorIsSupportedInWorker()` never resolves, so `MediaStreamVideoTrackSource._start()` never resolves, and `Output.start()` can stay pending forever.

We hit this in a production Next/Turbopack browser bundle while starting a WebCodecs recording: the recorder remained in its startup state because `Output.start()` never settled.

## Fix
The worker now handles an explicit `{ type: 'support' }` controller message. The main thread registers the listener first, then sends that request. This keeps the existing support cache and worker fallback behavior, but removes the startup-message race.

## Test plan
- npm run check

🤖 Generated with [Codex](https://Codex.com/Codex)

### DIFF
--- a/src/media-source.ts
+++ b/src/media-source.ts
@@ -2941,6 +2941,8 @@ type MediaStreamTrackProcessorWorkerMessage = {
 };
 
 type MediaStreamTrackProcessorControllerMessage = {
+	type: 'support';
+} | {
 	type: 'videoTrack';
 	trackId: number;
 	track: MediaStreamVideoTrack;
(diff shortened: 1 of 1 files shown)


Vanilagy:
The listener is registered synchronously after the creation of the Worker and thus the message will *always* arrive after the listener exists, so this is a non-fix.

Please submit the issue directly instead of a clanker pseudo-fix.

PR #387, opened 2026-05-25
Document UrlSource RequestInit overrides

samohovets:
## Summary
- Clarify in the Reading media files guide that `UrlSource` does not treat `requestInit` as a complete Fetch API pass-through.
- Document that Mediabunny overrides `signal` and `headers.Range` for both `requestInit` and `Request` inputs.
- Point readers to `Input.dispose` for canceling ongoing requests.

## Reasoning
The API docs already mention these overrides, but the higher-level Reading media files guide did not. Since users are likely to configure `UrlSource` from this guide, it should be clear there that `signal` and `Range` cannot be supplied through `RequestInit`/`Request` in the usual way.

## Testing
- Started the VitePress docs dev server locally and verified it runs.

### DIFF
--- a/docs/guide/reading-media-files.md
+++ b/docs/guide/reading-media-files.md
@@ -643,6 +643,8 @@ const source = new UrlSource('https://example.com/bigbuckbunny.mp4', {
 });
 ```
 
+All `requestInit` fields are respected except for `signal` and `headers.Range`, which are overridden by Mediabunny. The same applies to the `signal` and `headers.Range` values of a `Request` passed as the first constructor argument. To cancel ongoing requests, [dispose of the input](#disposing-inputs).
+
 `getRetryDelay` can be used to control the retry logic used should a request fail. When a request fails, `getRetryDelay` should return the time to wait in seconds before the request will be retried. Returning `null` prevents further retries.
 ```ts
 // UrlSource using retry logic with exponential backoff:


samohovets:
@Vanilagy as additional context: we’re currently adding `requestInit` support in Remotion’s `@remotion/media` integration mainly so users can control practical fetch options like `credentials` and `cache`, and customize headers for cases like authorization.

https://github.com/remotion-dev/remotion/pull/7493

 So from Remotion’s side, we’d also be happy to simply link to `UrlSource` docs and offload the exact semantics to Mediabunny. That makes it even more useful if the Mediabunny docs/types clearly state that this is not a fully transparent `fetch()` pass-through: Mediabunny owns `signal` and `Range` internally for cancellation and byte-range reads.

Maybe a follow-up could be to strip `signal` from the accepted `requestInit` type, so it’s obvious from TypeScript too?

Vanilagy (approved):
Goated, merged

Vanilagy:
The strip thing is a good idea, I'll do a few modifications in `main`

Vanilagy:
Peep this @samohovets 

https://github.com/Vanilagy/mediabunny/commit/c6c768d93ea83b148cb8adab3d7df0445f91829d

Vanilagy:
You could technically detect headers.Range on your end, remove it, and map it onto a call to `.slice()`

But thinking of it, maybe I should also just handle it directly! The reason it wasn't being handled is that in the old code, UrlSources *always* read from the start. After HLS, this was improved, and it can now start reading from any arbitrary offset.

samohovets:
@Vanilagy cool! Thanks 

Yeah the headers thing is not necessary. To not make the Typescript compiler go crazy. JSDoc is enough.

Vanilagy:
Peep the commit

PR #391, opened 2026-05-29
Fix tkhd issue for Apple ecosystem compatibility

StringWeaver:
# PR: fix tkhd issue for Apple ecosystem compatibility (align with FFmpeg)

## Description

When remuxing from formats (like some MKV files) where no track explicitly has its `disposition.default` set to `true` (or when the metadata is omitted), `mediabunny` currently passes this state directly to the output. As a result, the generated MP4 file may have all its tracks disabled (`tkhd` box flags are set to `0x02` instead of `0x03`).

While some tolerant players ignore this, strict players—particularly within the **Apple ecosystem (QuickTime Player, iOS/macOS Safari)**—require at least one video/audio track to have the `Track_enabled` flag (`0x01`). If this flag is missing, the player will fail to render the video (black screen) or play the audio.

This PR introduces a robust fallback logic during the `moov` box generation. If no track of a specific media type (video, audio, or subtitle) is explicitly enabled, it forces the first track of that type to be enabled.

### FFmpeg Alignment

This approach perfectly aligns with how FFmpeg handles the same scenario. FFmpeg checks the enabled track counts per media type before finalizing the header and applies a similar fallback to guarantee output compatibility:

```c
// libavformat/movenc.c (FFmpeg Source)
// Count enabled tracks per media type during muxing:
int enabled[AVMEDIA_TYPE_NB] = { 0 };
int first[AVMEDIA_TYPE_NB];

// ... (track iterations omitted) ...

for (i = 0; i < AVMEDIA_TYPE_NB; i++) {
    // If no track of a given type is enabled, force the first one to be enabled.
    if (!enabled[i] && first[i] >= 0)
        mov->tracks[first[i]].flags |= MOV_TRACK_ENABLED;
}
```

## Changes Made
- Modified `moov` box generation in `src/isobmff/isobmff-boxes.ts`.
- Added a pre-processing loop that checks `disposition.default` across `trackDatas`.
- If no track within a given type has `default !== false`, we mutate the `disposition` of the first track of that type to explicitly set `default = true`.
- This ensures the `tkhd` box generation (which checks for `disposition.default !== false`) will correctly apply the `0x01` (`Track_enabled`) flag.

## Testing

I have built this change locally and tested it in a PWA Media Player project ([StringWeaver/Local-PWA-Media-Player](https://github.com/StringWeaver/Local-PWA-Media-Player.git)). 
The resulting MP4 output can now be successfully imported and played on QuickTime and Safari without throwing the typical black screen error caused by the previously disabled `tkhd` flags.

### DIFF
--- a/src/isobmff/isobmff-boxes.ts
+++ b/src/isobmff/isobmff-boxes.ts
@@ -378,12 +378,26 @@ export const free = (size: number): Box => ({ type: 'free', size });
  * Movie Box: Used to specify the information that defines a movie - that is, the information that allows
  * an application to interpret the sample data that is stored elsewhere.
  */
-export const moov = (muxer: IsobmffMuxer) => box('moov', undefined, [
-	mvhd(muxer.creationTime, muxer.trackDatas),
-	...muxer.trackDatas.map(x => trak(x, muxer.creationTime)),
-	muxer.isFragmented ? mvex(muxer.trackDatas) : null,
-	udta(muxer),
-]);
+export const moov = (muxer: IsobmffMuxer) => {
+	// FFmpeg-aligned fallback: If no track of a given type is enabled, force the first one to be enabled.
+	// This ensures compatibility with strict players like QuickTime.
+	for (const type of ['video', 'audio', 'subtitle'] as const) {
+		const tracks = muxer.trackDatas.filter(t => t.type === type);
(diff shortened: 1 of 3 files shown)


Vanilagy:
Nice edge case catch. I moved the code around a little, this kinda logic belongs in the muxer, not in the box writer.

PR #394, opened 2026-05-30
Add standalone single-file HLS MP4 mode

alec-watts:
## Summary
- Add `singleFilePerPlaylistMode: 'standalone'` for HLS outputs that write one standalone fragmented MP4 media file.
- Preserve the existing single-file behavior as the default `segments` mode.
- Cover the new mode with a node test that verifies byte-range playlist output and standalone MP4 box layout/readability.

## Test plan
- `npx vitest run test/node/hls-output.test.ts`
- `npm run lint`
- `npm run build`
- `npm run test-node`

### DIFF
--- a/src/hls/hls-muxer.ts
+++ b/src/hls/hls-muxer.ts
@@ -32,6 +32,7 @@ import {
 	HlsOutputFormatOptions,
 	HlsOutputPlaylistInfo,
 	HlsOutputSegmentInfo,
+	Mp4OutputFormat,
 	OutputFormat,
 } from '../output-format';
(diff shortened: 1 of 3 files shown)


Vanilagy:
Thanks for the effort! Unfortunately, the approach wasn't quite sound. I implemented the feature in the attached commit.

In fact, there's zero new API! All you need to do now is use single-file mode with fMP4 segment format, and you'll automatically get a standalone-playable file out. It's a hardcoded special case, but I think it's fine here, and it mirrors FFmpeg I think too.

PR #397, opened 2026-06-01
Support QuickTime nclc color information

### LINKED ISSUE #398: Parse QuickTime nclc color information in ISOBMFF colr boxes
## Summary

Mediabunny currently parses ISOBMFF `colr` boxes when the color type is `nclx`, but QuickTime/MOV files can store the same numeric color primaries, transfer characteristics, and matrix coefficients using color type `nclc`.

When the demuxer ignores `nclc`, `InputVideoTrack.getColorSpace()` loses HDR metadata for affected MOV files. Downstream callers then cannot detect HDR sources from Mediabunny metadata alone.

## Expected behavior

For a MOV/ISOBMFF video track with a `colr` box whose color type is `nclc`, Mediabunny should read:

- color primaries
- transfer characteristics
- matrix coefficients

`fullRange` should remain unset for `nclc`, because unlike `nclx`, the QuickTime `nclc` payload does not include the full-range flag byte.

## Actual behavior

The ISOBMFF demuxer currently exits early unless the color type is `nclx`, leaving `colorSpace` unavailable for `nclc` sources.

## Related PR

I opened #397 with a small fix and a generated in-memory MOV test fixture that rewrites the `colr` box type from `nclx` to `nclc`.

Octember:
## Summary
- Fixes #398
- Parse ISOBMFF `colr` boxes with QuickTime `nclc` color type in addition to ISO `nclx`
- Preserve the `nclx` full-range flag behavior while leaving `fullRange` unset for `nclc`
- Fix `InputVideoTrack.hasHighDynamicRange()` to recognize PQ transfer as `pq`

## Test fixture note
The new node test creates a MOV in memory with normal `nclx` color metadata, then rewrites the color information box type from `nclx` to `nclc`. The helper targets the `colr` box payload directly, so the test exercises the demuxer path without adding a binary fixture.

## Tests
- `npx vitest --run test/node/read-mp4.test.ts`
- `npm run check`
- `npm run lint`
- `npm run test-node` currently fails locally in unrelated `node/server-extension.test.ts > Video > Memory doesn't leak` RSS assertion (`expected 12926976 to be less than 10485760` on isolated rerun; full suite saw `31309824`).

### DIFF
--- a/src/input-track.ts
+++ b/src/input-track.ts
@@ -691,7 +691,7 @@ export class InputVideoTrack extends InputTrack {
 		const colorSpace = await this._backing.getColorSpace();
 
 		return (colorSpace.primaries as string) === 'bt2020' || (colorSpace.primaries as string) === 'smpte432'
-			|| (colorSpace.transfer as string) === 'pg' || (colorSpace.transfer as string) === 'hlg'
+			|| (colorSpace.transfer as string) === 'pq' || (colorSpace.transfer as string) === 'hlg'
 			|| (colorSpace.matrix as string) === 'bt2020-ncl';
 	}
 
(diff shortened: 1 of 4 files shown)


Vanilagy (approved):

Vanilagy:
Thanks! Cleaned up the code a little, and actually made the ISOBMFF *muxer* write the `nclc` box when it's in QuickTime mode, which is the correct behavior.

PR #399, opened 2026-06-02
Fix HLS master playlist codec strings

soloish90:
## Summary

Fix HLS master playlist `CODECS` values so MediaBunny writes HLS/RFC-style codec identifiers (based on Apple hls specifications) instead of internal codec names.

This maps known HLS aliases such as MP3, AC-3, E-AC-3, and FLAC to the identifiers expected by HLS clients, while preserving already-complete codec strings like `avc1...`, `hvc1...`, `hev1...`, and `mp4a.40.*`.

## Details

- Adds an internal HLS codec-string serializer.
- Maps:
  - `aac` -> `mp4a.40.2`
  - `mp3` -> `mp4a.40.34`
  - `ac3` -> `ac-3`
  - `eac3` -> `ec-3`
  - `flac` -> `fLaC`
  - `webvtt` -> `wvtt`
- Avoids emitting incomplete parameter-dependent video names like `avc`, `hevc`, `vp9`, or `av1` when a full decoder config is unavailable.

## Tests

- `npx vitest --run test/node/hls-output.test.ts`
- `npx eslint src/hls/hls-codecs.ts src/hls/hls-muxer.ts test/node/hls-output.test.ts`
- `npx tsc -p src --noEmit`

### DIFF
--- a/src/hls/hls-codecs.ts
+++ b/src/hls/hls-codecs.ts
@@ -1,0 +1,40 @@
+/*!
+ * Copyright (c) 2026-present, Vanilagy and contributors
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+const HLS_CODEC_ALIASES = new Map([
+	['aac', 'mp4a.40.2'],
+	['mp3', 'mp4a.40.34'],
+	['mp4a.69', 'mp4a.40.34'],
+	['mp4a.6B', 'mp4a.40.34'],
+	['mp4a.6b', 'mp4a.40.34'],
+	['ac3', 'ac-3'],
+	['eac3', 'ec-3'],
+	['flac', 'fLaC'],
+	['webvtt', 'wvtt'],
+]);
+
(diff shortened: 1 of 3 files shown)


soloish90:
still needs more coverage for all codecs. working on that now (and doing some more testing of actual playback tests where possible). i mean, who has source files that cover all the bases, amiright? lol.

Vanilagy:
@soloish90 Please lead with the issue first instead of jumping straight to the PR. Mediabunny's HLS muxer is precisely written so that it emits the full codec strings, not just the internal identifiers!

Yes, `const codecString = trackData?.info.decoderConfig.codec ?? track.source._codec;` will default to the internal identifier *technically*, but only if no media data has been added for it, which is an extremely niche case.

Can you describe in more detail your setup and when/how you're hitting this?

soloish90:
yes, i see.  something i missed.  generally playing with more of a HLS JIT (just in time) VOD path...  in this case, i'm hitting the problem where master playlist  (and CODECS field) is generated first (as required)... the segments start generating on demand (and not sequentially necessarily).  i missed the the actual important thing about how MB conversion was generating the HLS.

so my bad, back to the drawing board. (simple fix: just grab my first media sample/segment before generating the master- as you already do. then i can concern myself with the on demand part.)

Issue #244, opened 2025-12-10
`outOfBandError` handling produces an unhandled rejection


JonnyBurger:
I think there is a problem in this code path here:

https://github.com/Vanilagy/mediabunny/blob/c0c353dbe8b3a3a4afddb55de734025df096b110/src/media-sink.ts#L406-L589

For reasons yet unknown, I am facing a decoding error only in CI, but this is not the problem I want to report.
It seems like there is "out of band" error handling which aims to produce a better error message than the default.

However, in my case it produces an error with no error message at all, and the original error remains unhandled.
Logs from [this](https://github.com/remotion-dev/remotion/actions/runs/20093290304/job/57645508149?pr=6040) CI run.

```
 FAIL   chromium  src/test/audio.test.tsx > should be able to render 2 audios
Unknown Error: Error
    at http://localhost:63315/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/deps/mediabunny.js?v=a29abe5f:4835:28
	# ^ this points to 
    at VideoSampleSink.mediaSamplesInRange (http://localhost:63315/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/deps/mediabunny.js?v=a29abe5f:4906:7)
    at VideoSampleSink.samples (http://localhost:63315/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/deps/mediabunny.js?v=a29abe5f:5568:17)
    at getFramesSinceKeyframe (http://localhost:63315/@fs/Users/runner/work/remotion/remotion/packages/media/dist/esm/index.mjs:2356:42)
    at async requestKeyframeBank (http://localhost:63315/@fs/Users/runner/work/remotion/remotion/packages/media/dist/esm/index.mjs:2543:26)
    at async extractFrameInternal (http://localhost:63315/@fs/Users/runner/work/remotion/remotion/packages/media/dist/esm/index.mjs:3052:24)
    at async Promise.all (index 0)
    at async extractFrameAndAudio (http://localhost:63315/@fs/Users/runner/work/remotion/remotion/packages/media/dist/esm/index.mjs:3132:28)
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
[...]

PR #405, opened 2026-06-11
Fix unhandled rejections from deliberate cancellations during dispose

tomyang11:
## Context

Same family as #197 (and the spirit of #244): when an `Input` is disposed mid-flight and the app then aborts its in-flight reads, several internal promises reject with nobody to hear them, and a **deliberate cancellation** surfaces in the console as an unhandled promise rejection.

I found three distinct routes by instrumenting `Promise.prototype.then` to attribute the rejecting promise's creation site (a multi-source composition player tearing down `Input`s while reads are in flight). I'm aware #203 was declined because blanket suppression would have masked the real bug there — these three are not suppressions of errors that have a consumer somewhere; in each case the promise is *consumer-less by design* or the rejection *is* the cancellation:

## Route 1 — `ReadOrchestrator.read()`, cache-satisfied branch

When the requested region is covered by cache but the prefetch region is not, `read()` returns the cached result, and the `promise` from `promiseWithResolvers` is never handed to the caller — it only tracks completion of the **speculative prefetch remainder** (the pending slice still references its `reject`). If the worker filling that remainder fails or its read is aborted during teardown, the rejection has no consumer.

Fix: attach a no-op handler in exactly that branch. A real failure of that region is re-encountered — with a consumer — by the next read that actually needs those bytes.

## Route 2 — `ReadOrchestrator.runWorker()`

The `catch` rethrows into a `void`'ed chain when a worker has no pending slices ("so it doesn't get swallowed"). For a **live** worker that's right. But `dispose()` flags workers `aborted` (and the orchestrator `disposed`) *before* the caller aborts the underlying reads — your own comment in `createWorker` documents the workers-after-disposal variant of this ordering. So a rejection arriving on an aborted worker / disposed orchestrator is the deliberate cancellation itself, carrying no information, and every speculative slice-free worker cancelled at teardown becomes an unhandled rejection.

Fix: rethrow only when `!worker.aborted && !this.disposed`. Real errors on live workers still propagate exactly as before.

## Route 3 — `Input.dispose()`

```ts
void this._demuxerPromise?.then(demuxer => demuxer.dispose());
```

This derived chain exists only to dispose the demuxer once it materializes. If `dispose()` is called while the demuxer is still initializing and the metadata read then fails/aborts, `_demuxerPromise` rejects — its real awaiters (`getPrimaryVideoTrack` etc.) receive the error properly, but this chain has no rejection handler. A demuxer that never came to exist needs no disposal.

Fix: `.catch(() => {})` on the derived chain only.

## Testing

- `npm run test-node`: 274 passed; `tsc -p src` and `eslint` clean
- Running in production as a package patch against 1.45.4: all three unhandled-rejection signatures are gone, error propagation to real awaiters unchanged

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -528,7 +528,13 @@ export class Input<S extends Source = Source> extends EventEmitter<InputEvents> 
 		this._sourceRefs.length = 0;
 
 		void this._demuxerPromise
-			?.then(demuxer => demuxer.dispose());
+			?.then(demuxer => demuxer.dispose())
+			// This derived chain exists only to dispose the demuxer once it materializes. If dispose() is
+			// called while the demuxer is still initializing (metadata read in flight) and that read then
+			// fails or is aborted, _demuxerPromise rejects - its real awaiters (getPrimaryVideoTrack etc.)
+			// receive the error, but this chain has no consumer and would surface the deliberate cancellation
+			// as an unhandled promise rejection. A demuxer that never came to exist needs no disposal.
+			.catch(() => {});
 	}
 
 	/**
--- a/src/source.ts
+++ b/src/source.ts
@@ -1988,7 +1988,13 @@ class ReadOrchestrator {
 				offset: innerStart,
 			} satisfies ReadResult));
 		} else {
-			// The requested region was satisfied by the cache, but the entire prefetch region was not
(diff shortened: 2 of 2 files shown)


Vanilagy:
Thanks, fixed it in a slightly different way in the attached commit. This PR was too comment-spammy to accept

PR #406, opened 2026-06-11
Add per-sink decoder preferences (hardwareAcceleration, optimizeForLatency)

tomyang11:
## What

Adds an optional `VideoSinkDecoderOptions` (`{ hardwareAcceleration?, optimizeForLatency? }`) second parameter to `VideoSampleSink`, exposed on `CanvasSink` as `options.decoderOptions`, applied to the decoder config before `VideoDecoderWrapper` is constructed. Validation mirrors `decode.ts`'s `validateVideoDecodingConfig`; the type is exported from the package root.

## Why

Applications that run **many sinks concurrently** (multi-track video editors / composition players) need to manage hardware decode sessions deliberately: the number of concurrent hardware sessions is OS-limited, undocumented, varies by hardware/codec/load, and exceeding it **fails silently** on some platforms — on macOS VideoToolbox the excess decoder accepts `configure()` and `decode()` and simply never outputs a frame. The standard mitigation is a decoder pool that places overflow sinks on `'prefer-software'` explicitly, which currently isn't expressible through `CanvasSink`/`VideoSampleSink`.

`optimizeForLatency` is exposed alongside it as the other WebCodecs decoder-config preference an application reasonably wants per sink (e.g. faster first-frame on seek-heavy UIs).

There's precedent for per-config injection in the codebase already: the interlaced-AVC Chromium workaround in `VideoDecoderWrapper` swaps to `'prefer-software'`. The new override composes cleanly with it — the workaround runs later and can only strengthen the preference toward software.

## Testing

- `npm run test-node`: 274 passed; `tsc -p src` and `eslint` clean
- Running in production (as a package patch against 1.45.4) in a composition player that pools ~6 hardware sinks and overflows the rest to `prefer-software` — the silent-stall failure mode is gone

Happy to adjust the API shape (naming, flat options vs. nested `decoderOptions`, validation placement) to your preference.

### DIFF
--- a/docs/guide/media-sinks.md
+++ b/docs/guide/media-sinks.md
@@ -225,6 +225,11 @@ Create the sink like so:
 import { VideoSampleSink } from 'mediabunny';
 
 const sink = new VideoSampleSink(videoTrack);
+
+// Optionally, configure the decoder:
+const sink = new VideoSampleSink(videoTrack, {
+	hardwarePreference: 'prefer-software',
+});
 ```
 
 #### Single retrieval
@@ -363,6 +368,8 @@ type CanvasSinkOptions = {
 	rotation?: 0 | 90 | 180 | 270;
 	crop?: { left: number; top: number; width: number; height: number };
 	poolSize?: number;
+	alpha?: boolean;
+	decoderOptions?: VideoSinkDecoderOptions;
 };
 ```
 - `width`\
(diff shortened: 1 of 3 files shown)


Vanilagy (approved):
Good change, long overdue. Simplified some of the docs.

PR #429, opened 2026-07-02
Fix duplicated entries in generated API docs

### LINKED ISSUE #428: API docs contain duplicate entries
The docs at https://mediabunny.dev/api/ contain duplicate entries. This also happens in the left sidebar

Image

pohzipohzi:
Fixes https://github.com/Vanilagy/mediabunny/issues/428

image

This commit is made in the same spirit as what is mentioned in the file header

```
// This script has been 100% vibe-coded with Claude (and Gemini!), meaning I literally haven't looked at any of the
// code. It's probably a mess, but it solves a one-off problem where only the output matters, and the output is indeed
// good, which is the point of a custom script for this: full, precise control.
```


Vanilagy (approved):
Usually anti vibe coding but this script embraces it 👍

PR #433, opened 2026-07-08
Add extension decoders to media player example

samohovets:
## Summary
- register the ProRes decoder extension in the media player example
- register the AC-3/E-AC-3 decoder extension in the media player example
- expand the file picker accept list to include all supported media container extensions
- add the missing Vite alias for @mediabunny/ac3

## Testing
- npm run build
- npm run examples:build

### DIFF
--- a/examples/media-player/media-player.ts
+++ b/examples/media-player/media-player.ts
@@ -738,32 +738,7 @@ window.addEventListener('resize', () => {
 selectMediaButton.addEventListener('click', () => {
 	const fileInput = document.createElement('input');
 	fileInput.type = 'file';
-	fileInput.accept = [
-		'video/*',
-		'audio/*',
-		'video/quicktime',
-		'video/x-matroska',
-		'video/webm',
-		'video/mp2t',
-		'application/ogg',
-		'application/vnd.apple.mpegurl',
-		'.mp4',
-		'.m4v',
-		'.m4a',
-		'.mov',
-		'.m4s',
-		'.mkv',
-		'.webm',
-		'.ogg',
-		'.ogv',
-		'.oga',
-		'.mp3',
-		'.wav',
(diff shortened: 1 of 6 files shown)


Vanilagy:
I've been wondering if I should do this but it's probably a good idea. I'd add a comment to the example code though, explaining why this is being done.

Can you also update the thumbnail generation, file generation, and HLS conversion examples? They all read input files.

samohovets:
@Vanilagy, it’s been very useful for me to test different videos in the official player example, as it lets me compare Remotion’s implementation with Mediabunny’s reference implementation.

 It was especially helpful for the recent ProRes issue with Remotion’s `<Video>` tag. I could verify that ProRes worked correctly in Remotion while using the Mediabunny player as a reference.

Would it also make sense to add a short note to the example page mentioning that it uses the custom ProRes and AC-3/E-AC-3 decoders, with links to their docs?

examples/media-player/media-player.ts
```diff
@@ -732,7 +738,32 @@ window.addEventListener('resize', () => {
...
+		'.mkv',
+		'.webm',
+		'.ogg',
+		'.ogv',
+		'.oga',
+		'.mp3',
+		'.wav',
+		'.aac',
+		'.flac',
+		'.ts',
```
Vanilagy: Why did you add this? What wasn't working? If this is necessary, it should be ported to the other files too.

Vanilagy (changes requested):

Vanilagy:
@samohovets I think it's not necessary to specify in the docs that the extensions are being used. I think it's pretty obvious when you look at the code, so it should be fine if somebody's confused and wants to figure it out.

Vanilagy (approved):

PR #434, opened 2026-07-08
feat: allow passing external audio source to Conversion API

unfoundbox:
This PR makes a small tweak to the `Conversion` pipeline so we can easily add audio to silent videos.

**use case**
We're building an app that adds synthesized AI voiceovers to screen recordings (which often don't have an initial audio track). Without this option, we had to abandon the nice `Conversion` API and manually wire up `EncodedPacketSink` and `Output` just to multiplex an external audio buffer. 

**approach**
- Added `externalAudioSource` to `ConversionOptions`.
- If provided, we inject it directly into the `Output` multiplexer at the end of the track initialization loop.
- This bypasses the `input.getPrimaryAudioTrack()` validation failures, letting the multiplexing happen seamlessly alongside the video!

edit - **trimming behavior**
Since an `externalAudioSource` is typically an `AudioSampleSource` where the developer pushes samples manually, this source currently bypasses the internal `ConversionOptions.trim` math. The developer remains responsible for only generating/pushing the audio samples that fit within their desired trim window.

I also added a quick interactive demo in `examples/external-audio` so you can see it in action (it generates a procedural beep and muxes it over a selected video).

Let me know if you'd prefer this configured differently. Happy to adjust!

### DIFF
--- a/examples/external-audio/external-audio.ts
+++ b/examples/external-audio/external-audio.ts
@@ -1,0 +1,100 @@
+import { 
+	Conversion, 
+	AudioSampleSource, 
+	AudioSample 
+} from '../../src/index';
+
+const selectFileBtn = document.getElementById('select-file') as HTMLButtonElement;
+const fileNameEl = document.getElementById('file-name') as HTMLParagraphElement;
+const loadingEl = document.getElementById('loading') as HTMLDivElement;
+const resultEl = document.getElementById('result') as HTMLDivElement;
+const outputVideo = document.getElementById('output-video') as HTMLVideoElement;
+const downloadLink = document.getElementById('download-link') as HTMLAnchorElement;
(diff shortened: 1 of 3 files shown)


Vanilagy:
Hey, I agree that this is a good feature for Mediabunny to support, and I think the conversion API at it stands today is slightly too limiting. The solution of this PR is a bit too direct and inflexible though, so I'll be closing this PR.

I think the correct shape for a modification of the conversion API would be the following: Add the ability to label a conversion as "non-owning", meaning it does not take full ownership of the output. Rather, it just adds tracks to the output and adds media data to those, but it doesn't finalize the output itself. Then, you can run a conversion for the track you want to copy, and then just manually add the new tracks to it, directly to the output. This should compose extremely well. 

I created a new issue #436 to track this. Let me know if you'd be interested in tackling this problem, but before you do, we should discuss API first.

Issue #436, opened 2026-07-10
"Unowning" Conversions


Vanilagy:
Conversions that don't "own" the output; they simply add tracks to the output, given an input, and drive their media data. Finalization etc. is still up to the user.

unfoundbox:
Hey, thanks for getting back. happy to take a crack at this. Here's the shape I landed on, prototyped on a branch to make sure it actually holds up.

One new option on `ConversionOptions`:

```ts
/**
 * Whether this conversion takes full ownership of the output, defaults to `true`. An owning conversion
 * requires a fresh output and controls its entire lifecycle: it starts it, writes its metadata tags, and
 * finalizes it. A non-owning conversion only adds tracks to the output and drives their media data;
 * starting and finalizing the output remain your responsibility.
 */
ownsOutput?: boolean;
```

With `ownsOutput: false`, a conversion just adds its tracks and pumps them during `execute()`. You own start/finalize/metadata, and can attach your own tracks or other non-owning conversions to the same output:

```ts
const output = new Output({ format: new Mp4OutputFormat(), target: new BufferTarget() });
const conversion = await Conversion.init({ input, output, ownsOutput: false });

const voiceover = new AudioBufferSource({ codec: 'aac', bitrate: 128e3 });
output.addAudioTrack(voiceover);

await output.start();
await Promise.all([
	conversion.execute(),
	(async () => {
		await voiceover.add(voiceoverBuffer);
		voiceover.close();
	})(),
]);
await output.finalize();
```

Composes for the two-file case too — video from A, audio from B — just run two non-owning conversions into the same output.

A few implementation decisions worth flagging:

- `execute()` throws if the output hasn't been started yet, instead of waiting for it - a forgotten `output.start()` should be a clear error, not a hang.
- `tags` throws when combined with `ownsOutput: false`. Two conversions both copying input tags would otherwise fight over `setMetadataTags`. You can still do `output.setMetadataTags(await input.getMetadataTags())` yourself.
- `cancel()` only closes the conversion's own sources, not the whole output - other tracks (yours or a sibling conversion's) keep going and `finalize()` still works. Means your own error handling needs to separately call `output.cancel()` if you want a full abort. A bit of a footgun, but seemed like the only sane behavior for something that doesn't own the output.
- Tracks default into `output.defaultTrackGroup` instead of getting their own pairability graph, so they pair with your tracks by default. Not sure this is right for HLS outputs, open to other opinions here.
[...]

PR #438, opened 2026-07-12
Add non-owning conversions via ConversionOptions.ownsOutput

unfoundbox:
## What this adds 

Context - https://github.com/Vanilagy/mediabunny/issues/436

`ownsOutput?: boolean` (default `true`, fully backwards-compatible). With `ownsOutput: false`, a `Conversion` only adds its own tracks to the output and pumps their media data during `execute()` — the caller owns the output lifecycle (additional tracks, metadata, `start()`, `finalize()`, `cancel()`). This cleanly models "a conversion is one of several contributors to a single output file."

Semantics:
- The output no longer needs to be fresh, but must still be `'pending'` at init (tracks can only be added before start).
- `execute()` throws with a descriptive message if the output hasn't been started.
- `conversion.cancel()` closes only the conversion's own sources, leaving the output alive for its other contributors; a full abort is `conversion.cancel()` + `output.cancel()`.
- `isValid` in this mode means "contributes ≥1 track"; format minimum-track-count rules are enforced by `output.start()` against the complete track set, as before.
- `tags` + `ownsOutput: false` throws at init — metadata belongs to the output's owner (`output.setMetadataTags()`).

Also includes 
- an `examples/external-audio` example 
- a fix releasing synchronizer waiters when canceling during output finalization
- a new section on the "Converting media files" guide page describing this mode.

Let me know if this needs any tweaks :)


Vanilagy:
Awesome, thanks! I went through and made some cleanups to the logic. I removed the example since it's too niche for a general example. Still need to go through the docs, but will do that another time. But then, I'll land this PR.

Vanilagy (approved):

PR #439, opened 2026-07-13
Account for multi-frame Opus packets when computing packet duration

medeiroz:
### The bug

`parseOpusTocByte` (`src/codec-data.ts`) only reads the `config` field of the Opus TOC byte and always assumes **one frame per packet**:

```ts
export const parseOpusTocByte = (packet: Uint8Array) => {
	const config = packet[0]! >> 3;

	return {
		durationInSamples: OPUS_FRAME_DURATION_TABLE[config]!,
	};
};
```

Per [RFC 6716, section 3.2](https://datatracker.ietf.org/doc/html/rfc6716#section-3.2), a packet may carry 1, 2 or an arbitrary number of frames. The frame count is encoded in the two low bits of the TOC byte (code 0 → 1 frame, code 1/2 → 2 frames, code 3 → the frame count byte that follows), and the packet's duration is the frame duration **times the frame count**.

The Ogg muxer derives its granule positions from this value (`src/ogg/ogg-misc.ts` → `extractSampleMetadata`), so for multi-frame packets the granule positions advance slower than the actual audio and the file declares a shorter duration than it really has.

### Why it matters in practice

Chromium's `MediaRecorder` packs **three 20 ms frames into each Opus packet**. Remuxing its WebM/Opus output to Ogg/Opus (`Conversion` + `OggOutputFormat`, no re-encoding) therefore yields a file that declares **a third of its real duration**. On a real 5.72 s microphone recording, the final granule position came out as 99840, i.e. 2.08 s at 48 kHz. The audio data is all there, but any consumer that trusts the container — speech-to-text services, `mutagen`, the `<audio>` element's duration and seek bar — truncates it. That is how we ran into this.

### The fix

Read the frame count code and multiply. Non-Ogg code paths are unaffected, since `parseOpusTocByte` is only used by `extractSampleMetadata`.

### Test

Added `test/node/ogg-muxer.test.ts`, which feeds the Ogg muxer synthetic Opus packets carrying three 20 ms frames each (TOC config 31, code 3) and asserts the final granule position. It fails on `main` (`expected 9600 to be 28800` — exactly one third) and passes with this change.

`npm run lint` is clean. On `npm run test-node`, the new test passes and no other test changed state. Note that a number of tests already fail on `main` on Windows, unrelated to this change: `new URL('.', import.meta.url).pathname` yields `/C:/...`, so `FilePathSource` receives `C:\C:\Users\...` and throws `ENOENT`.


Vanilagy (approved):
Thanks, I cleaned up the test slightly

PR #441, opened 2026-07-14
Fix WebVTT-in-MP4 output by starting the aux writer

hikashop-nicolas:
### Problem

`Mp4OutputFormat` / `CmafOutputFormat` list `webvtt` in their supported subtitle codecs, and the ISOBMFF muxer maps it to the `wvtt` sample entry, so muxing a WebVTT track into MP4/MOV looks supported. In practice it fails: `IsobmffMuxer.start()` starts the main writer but never calls `this.auxWriter.start()`. The aux writer is where subtitle sample boxes are assembled, so the first subtitle sample write trips its `started === false` assertion and muxing throws.

### Fix

Start the auxiliary writer alongside the main writer in `start()`. One line; no API or format change. Video/audio-only output is unaffected (the aux writer was already constructed, just never started).

### Repro

Create an `Output` with `Mp4OutputFormat`, add a `TextSubtitleSource('webvtt')`, add a cue, and `finalize()` — it asserts before this change and produces a valid `wvtt` track after.

Found while working on subtitle muxing; the fix is independent of any new codec.

### DIFF
--- a/src/isobmff/isobmff-muxer.ts
+++ b/src/isobmff/isobmff-muxer.ts
@@ -211,6 +211,10 @@ export class IsobmffMuxer extends Muxer {
 	async start() {
 		const release = await this.mutex.acquire();
 
+		// The auxiliary writer builds subtitle sample boxes in memory; it must be started too,
+		// else the first box write for any subtitle track asserts (started === false).
+		this.auxWriter.start();
+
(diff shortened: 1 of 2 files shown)


Vanilagy:
Thanks, good catch, added a test too

PR #448, opened 2026-07-28
Add custom quality factors and quantizer bitrate mode

### LINKED ISSUE #327: Custom Quality and quantizer support
Quality class type prohibits an arbitrary quality factor from being passed to the constructor.
For this reason only the predefined `QUALITY_*` values can be used.

This combined with the lack of `quantizer` support makes adjusting the quality a bit too rigid.

DonCarignan:
This makes `Quality`'s constructor public so an arbitrary factor can be passed instead of being limited to the five presets, and adds `'quantizer'` as a third `bitrateMode` for `avc`, `hevc`, `vp9` and `av1`. Closes #327.

In quantizer mode, `bitrate` becomes optional and is only used as a fallback, and a new `quantizer` option sets the value directly on the codec's own scale: 0-51 for avc and hevc, 0-63 for vp9, and 0-255 for av1, which uses a quantizer index. To stay codec-independent you can keep passing a `Quality` and a fitting quantizer gets derived for whichever codec you end up on, so `QUALITY_MEDIUM` becomes 26 on avc and 128 on av1. Per-frame quantizers work as well: `add()` accepts a `quantizer` in the encode options and routes it into the right codec dictionary.

lucadalli suggested putting the quantizer on `Quality` itself; I went with the bitrate mode instead, since `Quality` is shared with audio encoding where quantizers don't apply.

Not every browser knows the mode, and some reject the config outright instead of reporting it unsupported, so `isConfigSupported` is probed inside a try/catch and the encoder falls back to variable bitrate using a bitrate derived from the quantizer. `canEncodeVideo` reports on the mode too, returning false rather than throwing for values outside a codec's scale so that codec lists can still be probed.

I also plumbed the options through `Conversion`, since otherwise none of this is reachable from the API most people transcode with. That includes adding them to the `needsTranscode` check, without which setting only a bitrate mode quietly takes the stream-copy path and does nothing. One rule there is stricter than in the encoding config: because a conversion picks the codec for you, a raw `quantizer` requires you to name a `codec` too, otherwise the same number would mean very different things depending on what it landed on. Pass a `Quality` if you want the codec to stay open.

One thing you should know about before merging, which I've deliberately not handled here. On Apple platforms, quantizer mode for avc and hevc only ever runs on the hardware encoder (`prefer-software` reports those configs as unsupported), and low quantizers there aren't safe: QP 8 threw an `OperationError`, and sweeping QP 0-8 panicked the kernel inside the video encoder's IOMMU (dart-ave0, macOS 26.5.2, Chrome 150, M2 Pro, reproduced once). Everything from QP 16 up encoded fine across profiles. The docs note the hazard.

Written with AI assistance.


Vanilagy:
TODO FOR MYSELF

DON'T FORGET @MEDIABUNNY/SERVER SUPPORT

Vanilagy (approved):

PR #452, opened 2026-07-31
Add encoder-native video resizing

alec-watts:
## Summary
- add an opt-in `resizeMode: 'encoder'` transform that passes original-resolution frames to WebCodecs and configures `VideoEncoder` at the target dimensions
- support `fill` directly and `cover` by cropping in the source pixel format with `VideoFrame.copyTo()` before encoder scaling, avoiding the Canvas/RGB transform path
- preserve existing Canvas behavior by default and fall back to it for transforms that the encoder path cannot handle

## Motivation
Canvas 2D `drawImage()` downscaling can discard substantial detail before the frame reaches the encoder, particularly for screen recordings with text and sharp UI edges. Increasing encoder quality cannot recover detail already lost during that resize. Letting `VideoEncoder` scale the original-resolution decoded frame avoids the Canvas resampling step and produced materially sharper output in our matched browser tests, with improved SSIM and PSNR at the same encoder quality.

## Test plan
- [x] `npm run check`
- [x] ESLint on the changed source and browser test files
- [x] browser tests for conversion and media sources (50 passed)
- [x] verify encoder configuration uses target dimensions while callbacks receive original-resolution samples
- [x] decode a cover-scaled fixture and verify the expected center crop

### DIFF
--- a/src/conversion.ts
+++ b/src/conversion.ts
@@ -189,6 +189,12 @@ export type ConversionVideoOptions = {
 	 */
 	fit?: 'fill' | 'contain' | 'cover';
 	/**
+	 * How resizing is performed. `'canvas'` (default) resizes decoded frames before encoding and supports all fitting
+	 * modes. `'encoder'` passes frames at their original resolution and lets the encoder scale them, which may improve
+	 * quality or performance. Encoder resizing only supports the `'fill'` fitting mode.
+	 */
+	resizeMode?: 'canvas' | 'encoder';
+	/**
 	 * The angle in degrees to rotate the input video by, clockwise. Rotation is applied before cropping and resizing.
 	 * This rotation is _in addition to_ the natural rotation of the input video as specified in input file's metadata.
 	 */
@@ -378,6 +384,12 @@ const validateVideoOptions = (videoOptions: ConversionVideoOptions) => {
 	}
 	if (videoOptions?.fit !== undefined && !['fill', 'contain', 'cover'].includes(videoOptions.fit)) {
 		throw new TypeError('options.video.fit, when provided, must be one of \'fill\', \'contain\', or \'cover\'.');
+	}
+	if (videoOptions?.resizeMode !== undefined && !['canvas', 'encoder'].includes(videoOptions.resizeMode)) {
+		throw new TypeError('options.video.resizeMode, when provided, must be one of \'canvas\' or \'encoder\'.');
+	}
+	if (videoOptions?.resizeMode === 'encoder' && videoOptions.fit !== undefined && videoOptions.fit !== 'fill') {
+		throw new TypeError('options.video.fit must be \'fill\' when options.video.resizeMode is \'encoder\'.');
 	}
 	if (
 		videoOptions?.width !== undefined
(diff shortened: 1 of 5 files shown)


DonCarignan:
I'm AFK till next week but here are some agent review comments worth considering.

The premise checks out — the spec requires this ([`VideoEncoderConfig.width`](https://w3c.github.io/webcodecs/#dom-videoencoderconfig-width): "The encoder MUST scale any VideoFrame whose `[[visible width]]` differs from this value"), and Chrome scales with libyuv's box filter, which for large downscales beats the canvas path, especially since `VideoSample.transform()` currently draws at the default `imageSmoothingQuality` of `'low'`. That also suggests a cheap complementary fix: setting it to `'high'` would improve all fit modes on all browsers, and would lift the fallback paths this PR itself relies on.

The custom encoder skip breaks more than it looks like, though. In Node there's no global `VideoEncoder`, so with `resizeMode: 'encoder'` the whole of `@mediabunny/server` errors out for every codec with "VideoEncoder is not supported by this browser." And since `canEncodeVideo`/`getFirstEncodableVideoCodec` still consult custom encoders, a `Conversion` will select a codec and then throw mid-execution. I'd either fall back to canvas resizing when only a custom encoder matches, or let custom encoders opt in via a static flag — `NodeAvVideoEncoder` already carries a swscale context, so it could support this directly. At minimum it needs a clear validation error instead of the misleading "not supported by this browser." As an aside, this also rules out ProRes entirely: no browser encodes it natively, so custom encoders are the only route it has.

`getNativeCrop` has an alignment crash: `copyTo`'s rect x/y must be sample-aligned for subsampled formats (the spec's "Verify Rect Offset Alignment" throws a `TypeError` otherwise), and `Math.round((width - newWidth) / 2)` produces odd offsets. An 854×480 source covered to a square target gives `left = 187`, which throws on every I420/NV12 frame. left/top (and ideally the dimensions too) should be rounded to multiples of 2.

Why copy at all? Scaling is keyed to the frame's *visible* dimensions, so `new VideoFrame(frame, { visibleRect })` should perform the cover crop with no pixel copy — no readback for GPU-backed frames, no per-frame `ArrayBuffer`, and no `assert(videoFrame.format)`, which currently crashes on the `format === null` hardware frames from #352. Is there a reason that doesn't work?

`'cover'` also silently becomes a stretch on the fallback paths: whenever `force` is set (rotation, non-square pixels, or the conversion probe failing), `getNativeCrop` bails and the canvas transform runs without target dimensions, so the fit is never applied anywhere and the encoder just fills — rotated video with `'cover'` comes out squashed. Related: when the probe fails, the catch keeps `resizeMode: 'encoder'` active, so if the probe failed because of the size mismatch itself, the "fallback" fails identically. Clearing `resizeMode` there would turn the probe into a real capability gate — which matters because Safari currently has no green signal for encoder scaling at all (the [WPT rescaling suite](https://wpt.fyi/results/webcodecs/video-encoder-rescaling.https.any.html) fails across the board there, masked by an unrelated `copyTo` bug), and the browser tests here only run Chrome.

Two smaller things: the encoder path drops `sample.rotation` and the PAR normalization that `transform()` performs, which is observable when using `VideoSampleSource` directly (Conversion's rerender conditions happen to cover it) — worth handling or documenting as a restriction. And the resize behavior is described in the docs in a few places (converting-media-files, media-sources) that would need updating.

None of this is fatal to the idea — the `'fill'` path especially is a clear win — but the custom encoder story needs an answer before this is safe to expose.

Vanilagy:
I wasn't actually aware that you could use VideoEncoder to do resizing, cool! However, I looked through the Chromium source code and it seems force a CPU readback path in all cases, since it does the actual resizing using libyuv.

I think a much simpler approach that also remains fully on the GPU is simply this:
https://github.com/Vanilagy/mediabunny/commit/a0e18ffde057a0f6c316903390f4bebf1186a58f

Released it in v1.52.2. Can you try it out @alec-watts?

DonCarignan:
@alec-watts heads-up: the alignment issue from the review above turned out to be real in practice — we hit it running a backport of this branch. Rotation-free portrait sources crash deterministically: 1080×1920 covered to an auto-deduced 406×720 target produces a cover crop of `{y: 3, height: 1915}`, and the odd values throw a `TypeError` at `copyTo()` on every I420/NV12 frame. Flooring `left`/`top`/`width`/`height` to even after the cover math fixes it; worth folding into this PR. While you're in there, the `assert(videoFrame.format)` path deserves a graceful fallback to the uncropped frame too — `format` can be null on some hardware-decoded frames, and a failed crop currently kills the whole encode instead of degrading to a crop that's off by a pixel.

Vanilagy:
Closing, works well in main now

PR #458, opened 2026-08-06
UrlSource: honor explicitly-set maxCacheSize in the no-range fallback

### LINKED ISSUE #457: UrlSource: falling back from range requests sets maxCacheSize = Infinity
In `src/source.ts`, when a server answers a `Range` request with something other than `206`, `UrlSource` falls back to streaming the whole file and disables its cache bound:

```js
worker.currentPos = 0;
this._orchestrator.options.maxCacheSize = Infinity; // 🤷
```

The `🤷` suggests this is a known compromise, but the consequence is that a caller who explicitly passed `maxCacheSize` to bound memory silently loses that bound based on a server response they do not control. For a large input the source will then buffer without limit — with a multi-GB file that is not a degraded experience so much as an OOM.

The fallback itself makes sense; it is the unbounded cache that seems surprising. A couple of directions:

- Keep the caller's `maxCacheSize` and evict as usual. Sequential streaming is exactly the access pattern a bounded cache handles well; the cost would be re-reading on backward seeks.
- Or keep `Infinity` as the default fallback but honour an explicitly-passed `maxCacheSize`, so opting into a bound is respected.
- Failing either, surfacing it more loudly than a console warning about range support would help — the memory consequence is not obvious from the warning text.

Context: found while tracking down memory growth in long multi-source renders through `@remotion/media`, which builds a `UrlSource` per distinct `src`. Not the cause of the problem I was chasing (my origin does support ranges), but it looked worth reporting.

Yasirdora:
Fixes #457.

When the server doesn't answer a Range request with 206, UrlSource falls back to streaming the entire resource through one worker and previously forced `maxCacheSize = Infinity` — silently discarding a memory bound the caller explicitly opted into, which can mean OOM instead of degraded performance for multi-GB inputs.

**Change:** the fallback now only lifts the cache bound when `maxCacheSize` was *not* explicitly provided. Default behavior is unchanged: with an unbounded cache, backward seeks never trigger a full re-download (which is why the bound was lifted in the first place — with a bounded cache, a backward seek past the eviction boundary forces re-downloading the entire resource, since the server ignores range requests). A caller who explicitly sets `maxCacheSize` gets their memory budget honored and accepts that trade-off — this is option 2 from the issue.

Also documents the behavior in the `maxCacheSize` docblock.

**Tests:** two new cases in `test/browser/url-source.test.ts` using a `fetchFn` stub that always responds `200` (ignoring `Range`) and streams a 16 KiB resource in 1 KiB chunks:

- explicit `maxCacheSize` is preserved through the fallback, sequential reads evict as usual, and a backward read to evicted data still returns correct bytes (via one additional fetch)
- without an explicit bound, the cache still becomes unbounded and backward reads are served from cache with no additional request

Verified: `npm run check`, ESLint on changed files, and the new test logic executed against `src/` directly. I couldn't run the headed-browser suite locally, so the browser tests should run in CI.

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -697,7 +697,12 @@ export type UrlSourceOptions = {
 	 */
 	getRetryDelay?: (previousAttempts: number, error: unknown, url: string | URL | Request) => number | null;
 
-	/** The maximum number of bytes the cache is allowed to hold in memory. Defaults to 64 MiB. */
+	/**
+	 * The maximum number of bytes the cache is allowed to hold in memory. Defaults to 64 MiB.
+	 *
+	 * This bound is always honored, even when the server doesn't support range requests. Note that in that case,
+	 * a backward seek to already-evicted data requires re-downloading the entire resource.
+	 */
 	maxCacheSize?: number;
 
 	/** The maximum number of parallel requests to use for fetching. Defaults to 2. */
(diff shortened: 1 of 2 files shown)


Vanilagy:
Thanks for the PR, but I'm already working on a fix with a very different "shape" than this one.

Yasirdora:
Your approach with the sequential ReadableStreamSource backing is much cleaner. Thanks for the quick fix!

Vanilagy:
Thanks! What compelled you to file a PR?

Yasirdora:
I'm actaully building [nyhmas](https://github.com/Yasirdora/nyhmas), a client-side audio visualizer — WebGL effects with a deterministic MP4 export path, where mediabunny handles the muxing. It's a core dependency, so when I saw #457 pop up I figured it was a well-scoped way to get familiar with the codebase and give something back. :)

Vanilagy:
Awesome, thanks! Good luck with your project, and thank you for using Mediabunny.

PR #464, opened 2026-08-17
docs: note that trimming the start forces a transcode

karuha-m:
While building a browser-based trimming tool on Mediabunny, I ran into a performance cliff that the docs don't currently mention: `trim: { end }` stays on the fast copy path, but `trim: { start }` re-encodes the whole remaining file.

It's clear from the source once you look — in `conversion.ts`, `needsTranscode` includes `firstTimestamp < this._startTimestamp` for video, and the audio fast path is gated on the equivalent `needsTrimming` check. But it isn't discoverable from the docs: the "Transcoding video" / "Transcoding audio" sections enumerate what forces a transcode (`quality`, `keyFrameInterval`, `forceTranscode`, …) and `trim` isn't among them, and the "Trimming" section doesn't mention cost at all. The practical difference is large — trimming 10s off the front of a one-hour video re-encodes 59 minutes, while trimming from the end is near-instant.

This PR is docs only:

- **Trimming**: adds a short note contrasting the cost of `start` vs `end`, plus a line that negative `start` values don't trigger it (they lie before the track's first timestamp, so the condition is false).
- **Transcoding video / Transcoding audio**: one cross-reference line each, so `trim` shows up alongside the other transcode-forcing options.

No claim about *why* it's implemented this way — happy to reword if you'd rather state the reason, or to drop the cross-references if you feel they're redundant.

Context: I maintain [VideoKit](https://www.videokit.cc/en), which uses Mediabunny for in-browser conversion — this is one of the findings from the [Show and tell thread](https://github.com/Vanilagy/mediabunny/discussions/460), where I offered to turn them into a docs PR.

### DIFF
--- a/docs/guide/converting-media-files.md
+++ b/docs/guide/converting-media-files.md
@@ -238,6 +238,8 @@ Use the `quality` property to control the quality of the output video. For examp
 Use the `keyFrameInterval` property to control the maximum interval in seconds between key frames in the output video. Setting this fields forces a transcode.
 If you want to prevent direct copying of media data and force a transcoding step, use `forceTranscode: true`.
 
+Note that [trimming the start](#trimming) of a track also forces a transcode.
+
 Use the `hardwareAcceleration` property to control whether hardware or software acceleration is used for video transcoding.
 
 ### Processing video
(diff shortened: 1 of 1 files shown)


Vanilagy:
@karuha-m Thanks, I worded it more tersely and only in a single place in the linked commit.

PR #465, opened 2026-08-17
Guard decoder close against already-closed codecs (completes the #446 fix)

### LINKED ISSUE #446: VideoSampleSink/AudioSampleSink pumps never close their decoder when the pump body throws; ReadableStreamSource._dispose() drops pending reads un-settled
**Version:** 1.51.0 (npm `latest`). `src/media-sink.ts` and `src/source.ts` are byte-identical on `main` @ `4b5ef55`, so this is current. Chrome 150 / macOS; HEVC fMP4; reproduced identically on the main thread and in a module Worker.

Two independent bugs. They compound, but each needs its own fix.

### Bug 1 — `decoder.close()` is only on the happy path

In `mediaSamplesInRange`, the pump IIFE calls `decoder.close()` at **media-sink.ts:572**, after `await packets.next()` (559), `await packets.return()` (567) and `await decoder.flush()` (570). If any of those throws, control jumps to the IIFE's `.catch()` at **580**, which records the error into `outOfBandError` and **never closes the decoder**. `mediaSamplesAtTimestamps` has the same shape: close at **799**, catch at **803**.

This is not exotic — any read error during iteration hits it. Minimal repro, public API only, **no `dispose()` involved**:

```js
import { Input, Mp4InputFormat, CustomSource, VideoSampleSink } from 'mediabunny';

const configured = new Set(), closed = new Set();
const oc = VideoDecoder.prototype.configure, ocl = VideoDecoder.prototype.close;
VideoDecoder.prototype.configure = function (...a) { configured.add(this); return oc.apply(this, a); };
VideoDecoder.prototype.close     = function (...a) { closed.add(this);     return ocl.apply(this, a); };

const BYTES = new Uint8Array(await (await fetch('clip.mp4')).arrayBuffer());
let fail = false;
const source = new CustomSource({
  getSize: () => BYTES.length,
  read: async (start, end) => {
    await new Promise(r => setTimeout(r, 20));
    if (fail) throw new Error('simulated network failure');
    return BYTES.subarray(start, end);
  },
});

const input = new Input({ formats: [new Mp4InputFormat()], source });
const sink = new VideoSampleSink(await input.getPrimaryVideoTrack());

let n = 0;
try {
  for await (const s of sink.samples()) { s.close(); if (++n === 5) fail = true; }
} catch (e) { console.log('consumer saw:', e.message); }

await new Promise(r => setTimeout(r, 2000));
console.log([...configured].map(d => d.state)); // ['configured'] — expected ['closed']
```

Observed: `consumer saw: simulated network failure` then `['configured']`. Same with `sink.samplesAtTimestamps(...)` (the 799/803 pump). Same if instead of a read failure you do `break` + `input.dispose()` while reads are in flight — after `dispose()` the demuxer's next `requestSlice` throws `InputDisposedError` (`reader.ts:31`), which unwinds the pump into the same catch.

Repeating the cycle 30× (holding only `WeakRef`s) gives 30 configured / 0 closed decoders alive simultaneously. GC does eventually reclaim them — after memory pressure all 30 finalized — so it isn't unbounded, but until then every recycle leaves a live decoder plus a parked pump/demuxer/source cache.

This also contradicts two documented contracts: `Input.dispose()` says *"any open decoders will be closed"* (`docs/guide/reading-media-files.md`), and `break` *"will automatically clean up any internal resources (such as decoders)"* (`docs/guide/media-sinks.md`). Neither holds on the error path. `break` on its own, with no error and no dispose, does close the decoder correctly.

**Suggested fix:** *move* `decoder.close()` from 572 / 799 into a `finally` around the pump body. It has to be moved rather than added — `DecoderWrapper.close()` (media-sink.ts:1289) calls `this.decoder.close()` unguarded, so a second call would throw `InvalidStateError`.

### Bug 2 — `ReadableStreamSource._dispose()` drops pending reads un-settled

`source.ts:1667`:
```ts
_dispose() {
  this._pendingSlices.length = 0;   // <- 1668: promises dropped, never settled
  this._cache.length = 0;
  void this._reader?.cancel();
}
```
Truncating happens *before* `cancel()`, so when `_pull()`'s `await this._reader.read()` resolves `done`, the `pendingSlice.resolve(null)` loop at 1599-1602 has nothing left to resolve. Anything parked on such a read never resumes: the packet pump stays suspended inside `getNextPacket`, `onQueueNotEmpty()` never fires, and a sample pump parked at `await packets.next()` (559) — which is itself parked at `await queueNotEmpty` (403) — hangs forever. A consumer that keeps awaiting `next()` instead of returning parks at **612** and hangs too.

```js
import { Input, Mp4InputFormat, ReadableStreamSource, VideoSampleSink } from 'mediabunny';

const created = [];
const Native = globalThis.VideoDecoder;
globalThis.VideoDecoder = class extends Native {
  constructor(...args) { super(...args); created.push(this); }
};

const bytes = new Uint8Array(await (await fetch('clip.mp4')).arrayBuffer());
let offset = 0;
const stream = new ReadableStream({
  async pull(controller) {
    await new Promise(r => setTimeout(r, 100));           // slow producer: 64 KiB / 100 ms
    if (offset >= bytes.length) return controller.close();
    controller.enqueue(bytes.subarray(offset, offset += 65536));
  },
});

const input = new Input({
  formats: [new Mp4InputFormat()],
  source: new ReadableStreamSource(stream),
});
const sink = new VideoSampleSink(await input.getPrimaryVideoTrack());

const it = sink.samples()[Symbol.asyncIterator]();
for (let i = 0; i < 5; i++) { const { value } = await it.next(); value?.close(); }

void it.return();   // same as `break`
input.dispose();

await new Promise(r => setTimeout(r, 1000));
console.log(created.map(d => d.state));   // ['configured'] — expected ['closed']
```

Output: `['configured']`, 3/3 runs, main thread and Worker. Instrumenting the source shows exactly one `_read()` promise left un-settled, and instrumenting the `packets()` iterator shows the pump's last `next()` never settling. It reproduces whenever a source read is in flight at `dispose()` time; with an instant-delivery `ReadableStream` the same script tears down cleanly, so it is timing-dependent, not universal. Awaiting `it.return()` before `dispose()` makes no difference.

**Suggested fix:** reject the pending slices in `_dispose()` (e.g. with `InputDisposedError`) instead of dropping them — the same thing #404 fixed for `ReadOrchestrator`'s queued reads.

### The two fixes are not interchangeable

Patching only Bug 2 (making `_dispose()` reject) on the script above makes the pump's `packets.next()` settle — it rejects with `InputDisposedError` — but the decoder is **still** left `configured`, because the rejection lands in the catch at 580. Conversely, a `finally` alone does not fix Bug 2, since with the pending read dropped the pump body never completes and the `finally` never runs.

(Possibly related, not verified: `ReadOrchestrator.dispose()` at `source.ts:2448` also clears `workers` without settling their pending slices. I couldn't get a source on that path into that state — my `CustomSource` reads all settled — but it looks like the same shape as #404.)

---

Context: hit in a 24/7 video-wall product that recycles live HEVC tiles; each recycle left a configured hardware decoder plus a parked pump until GC. Happy to test a fix branch against our reproduction harness.

unfoundbox:
`90ec401` (fixes #446) moved `decoder.close()` into a `.finally()` around both
sink pumps. The codec may already be closed by the time that runs, and
`DecoderWrapper.close()` closes it unguarded.

When the platform fails a decoder,
[Close VideoDecoder](https://www.w3.org/TR/webcodecs/) sets `[[state]]` to
`"closed"` at step 2 and invokes the error callback at step 4. So:

1. `error:` fires at `media-sink.ts:972`, records `outOfBandError`, doesn't throw.
2. The pump's next `decode()`/`flush()` rejects against the closed codec.
3. `.finally()` at `:814` (and `:587`) calls `decoder?.close()`.
4. `VideoDecoderWrapper.close()` at `:1300` closes an already-closed codec →
   `InvalidStateError`.

That throw lands in a promise chain nobody holds, so it becomes an unhandled
rejection no consumer can catch — not with `try/catch` around iteration, not
through the sink's own error plumbing. `AudioDecoderWrapper.close()` at `:2205`
is the same shape.

#446 predicted it:

> It has to be moved rather than added — `DecoderWrapper.close()` calls
> `this.decoder.close()` unguarded, so a second call would throw
> `InvalidStateError`.

Moving it stops Mediabunny double-closing its own codec, but not the platform
closing it first.

Affects 1.52.0 through 1.54.0.

## Fix

The guard you already use on the encoder side (`media-source.ts:979-983`) and on
the decoder side in `v2` (`decode.ts`):

- `VideoDecoderWrapper.close()` — check `state !== 'closed'` for `decoder` and
  `alphaDecoder`, and make the custom-decoder path idempotent.
- `AudioDecoderWrapper.close()` — same.

Nothing changes where the codec is still open.

## How we hit it

`CanvasSink.getCanvas(t)` driven concurrently past the machine's hardware decoder
ceiling. Measured with `@motionvector/webcodecs-census`, our WebCodecs lifecycle
instrumentation, over CDP — Chrome for Testing 152, two 1080p H.264 clips:

| | |
| --- | --- |
| decoders constructed in 7.5s | 1233 |
| platform decode errors | 402 |
| `close()` on an already-closed codec | 148 |
| unhandled rejections | 175 |

The 1233 decoders were our bug, and bounding concurrency fixed our side. The
uncatchable rejection is reachable from any decode error, and it's the part a
consumer can't work around.

## Test

`test/browser/media-sinks.test.ts` stubs `VideoDecoder` so a decode closes the
codec and then invokes the error callback, matching the spec order. It asserts
the consumer still sees the failure through the iterator, and that nothing else
escapes.

Fails without the guard, passes with it. `expect(caught).toBeInstanceOf(DOMException)`
holds in both arms, so consumers observe no change.

`npm run lint`, `npm run check` and `npm run test-node` (360/360) clean.

### DIFF
--- a/src/media-sink.ts
+++ b/src/media-sink.ts
@@ -874,6 +874,7 @@ class VideoDecoderWrapper extends DecoderWrapper<VideoSample> {
 
 	customDecoder: CustomVideoDecoder | null = null;
 	customDecoderCallSerializer = new CallSerializer();
+	customDecoderClosed = false;
 	customDecoderQueueSize = 0;
 
 	inputTimestamps: number[] = []; // Timestamps input into the decoder, sorted.
@@ -1294,11 +1295,21 @@ class VideoDecoderWrapper extends DecoderWrapper<VideoSample> {
 
 	close() {
 		if (this.customDecoder) {
-			void this.customDecoderCallSerializer.call(() => this.customDecoder!.close());
+			if (!this.customDecoderClosed) {
+				this.customDecoderClosed = true;
+				void this.customDecoderCallSerializer.call(() => this.customDecoder!.close());
+			}
 		} else {
 			assert(this.decoder);
-			this.decoder.close();
-			this.alphaDecoder?.close();
+
+			// The codec may already be closed: when the platform fails a decoder, the spec closes it before invoking
+			// the error callback, and close() on a closed codec throws InvalidStateError.
+			if (this.decoder.state !== 'closed') {
+				this.decoder.close();
+			}
+			if (this.alphaDecoder && this.alphaDecoder.state !== 'closed') {
+				this.alphaDecoder.close();
+			}
 
 			this.colorQueue.forEach(x => x.close());
 			this.colorQueue.length = 0;
@@ -2065,6 +2076,7 @@ class AudioDecoderWrapper extends DecoderWrapper<AudioSample> {
 
 	customDecoder: CustomAudioDecoder | null = null;
 	customDecoderCallSerializer = new CallSerializer();
+	customDecoderClosed = false;
 	customDecoderQueueSize = 0;
 
 	// Internal state to accumulate a precise current timestamp based on audio durations, not the (potentially
@@ -2199,10 +2211,17 @@ class AudioDecoderWrapper extends DecoderWrapper<AudioSample> {
 
 	close() {
 		if (this.customDecoder) {
-			void this.customDecoderCallSerializer.call(() => this.customDecoder!.close());
+			if (!this.customDecoderClosed) {
+				this.customDecoderClosed = true;
+				void this.customDecoderCallSerializer.call(() => this.customDecoder!.close());
+			}
 		} else {
 			assert(this.decoder);
-			this.decoder.close();
+
+			// See the comment in VideoDecoderWrapper.close().
+			if (this.decoder.state !== 'closed') {
+				this.decoder.close();
+			}
 		}
 	}
 }
--- a/test/browser/media-sinks.test.ts
+++ b/test/browser/media-sinks.test.ts
@@ -3,7 +3,7 @@ import { Input } from '../../src/input.js';
 import { UrlSource } from '../../src/source.js';
 import { ALL_FORMATS } from '../../src/input-format.js';
 import { assert } from '../../src/misc.js';
-import { AudioSampleSink } from '../../src/media-sink.js';
+import { AudioSampleSink, VideoSampleSink } from '../../src/media-sink.js';
 
 // https://github.com/Vanilagy/mediabunny/issues/370
 test('Negative audio timestamps are preserved', async () => {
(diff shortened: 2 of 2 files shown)


Vanilagy:
Already fixed in https://github.com/Vanilagy/mediabunny/pull/462. The logic with the custom decoder is also unnecessary

PR #468, opened 2026-08-20
Support missing or empty avcC chunk

matanui159:
Sometimes FFmpeg will generate missing or empty `avcC` chunks if the stream doesn't have a global header. This adds support for those files in the same way mediabunny already supported matrsoka files without a `CodecPrivate` element.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ node_modules
 /dist
 /dist-docs
 .DS_Store
-/docs/.vitepress/cache
+.vitepress/cache
 /docs/api
 *.tsbuildinfo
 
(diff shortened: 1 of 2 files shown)


Vanilagy:
Thanks, could you send me one such file? If it can't be shared publicly, send it to me on Discord (`@vanilagy`)!

matanui159:
Sorry for the delay, starting to find out that this issue might be very specific to the version and build configuration of our build of FFmpeg that we use. Happy to still go through with the "fix" if you are though, here's an example file (converted from the existing `annex-b-avc.mkv` file):
https://github.com/user-attachments/assets/9e0cc9da-84f6-4232-8789-c835e3daec93

The file works with FFmpeg and also Chrome. But struggles in Quicktime and Safari. Haven't tested any other video players.

Vanilagy:
Thanks for the fix and your PR! It was exactly right, I just moved the code around a little so the comment could refer solely to the "box being empty" case. The file now plays perfectly in Mediabunny! <3

Deploying the fix in the next version.

PR #470, opened 2026-08-23
Fix tenc fallback for encrypted ISOBMFF files

### LINKED ISSUE #469: ISOBMFF demuxer fails on encrypted files with empty/non-existent senc
I actually ran into this bug when implementing a dash demuxer, but the bug applies to existing isobmff support too.

According to ISO/IEC 23001-7, 

- A sample_count of 0 in a senc box is valid (see repro)

> sample_count is the number of CENC SAI coded in the SampleEncryptionBox. For version 0, it shall be
either 0 or the number of samples in the track or track fragment where the SampleEncryptionBox
is contained.

- A senc box is non-mandatory

For both of these cases (empry senc or no senc) mediabunny currently just assumes no encryption. This is wrong and it should fall back to the encryption/iv specified by the tenc box.

The file I came across seemed to have been created by AWS Elemental MediaPackage so maybe all files from it are like this.

## Repro

KID=00112233445566778899aabbccddeeff
KEY=aabbccddeeff00112233445566778899

File:
https://github.com/user-attachments/assets/b752f7ad-7ce8-43c2-8ae9-88e74c6ae696

working shaka decrypt command:
```bash
shaka-packager \
  "in=bug_empty_senc.mp4,stream=audio,output=dec.mp4" \
  --enable_raw_key_decryption \
  --keys=label=:key_id=00112233445566778899aabbccddeeff:key=aabbccddeeff00112233445566778899
```

jepcd:
Fixes https://github.com/Vanilagy/mediabunny/issues/469


Vanilagy:
Thank you for the clean fix! <3

Also I totally fell for that rick roll goddamnit 😩

PR #475, opened 2026-08-26
Use AVC SPS color space when ISOBMFF colr is missing

### LINKED ISSUE #474: Bad parsing of colorSpace in SPS when `colr` atom is missing
Here is a good writeup of the bug, I think it's best to just link it: https://github.com/remotion-dev/remotion/issues/10841

I've managed to isolate this (Bug is arguably already at `getColorSpace()`, but shows nicely how it trickles down to WebCodecs having an issue):

```
<script type="module">
	import {
		ALL_FORMATS,
		Input,
		UrlSource,
		VideoSampleSink,
	} from 'https://cdn.jsdelivr.net/npm/mediabunny@1.55.1/dist/bundles/mediabunny.min.mjs';

	const src =
		'https://remotion.media/issue-10841-full-range-bt470bg-no-colr.mp4';

	const reproduce = async () => {
		const input = new Input({
			formats: ALL_FORMATS,
			source: new UrlSource(src),
		});

		try {
			const videoTrack = await input.getPrimaryVideoTrack();
			if (!videoTrack) {
				throw new Error('Expected the input to have a video track.');
			}

			const trackColorSpace = await videoTrack.getColorSpace();
			const decoderConfig = await videoTrack.getDecoderConfig();
			if (!decoderConfig) {
				throw new Error('Expected a VideoDecoderConfig.');
			}

			const sink = new VideoSampleSink(videoTrack);
			const sample = await sink.getSample(0);
			if (!sample) {
				throw new Error('Expected a decoded video sample.');
			}

			const decodedColorSpace = sample.colorSpace.toJSON();
			sample.close();

			const result = {
				src,
				expectedFromSps: {
					matrix: 'bt470bg',
					fullRange: true,
				},
				trackColorSpace,
				decoderConfigColorSpace: decoderConfig.colorSpace,
				decodedColorSpace,
			};

			globalThis.reproductionResult = result;
			console.log('Issue #10841 reproduction result:', result);

			if (
				decoderConfig.colorSpace === undefined &&
				decodedColorSpace.matrix === 'bt709' &&
				decodedColorSpace.fullRange === false
			) {
				console.error(
					'REPRODUCED: SPS/VUI declares full-range BT.470BG, but WebCodecs decoded limited-range BT.709.',
				);
			} else {
				console.warn(
					'NOT REPRODUCED on this browser. Inspect the result above; decoder behavior may be platform-specific.',
				);
			}

			return result;
		} finally {
			input.dispose();
		}
	};

	globalThis.reproduction = reproduce().catch((error) => {
		globalThis.reproductionError = error;
		console.error('Issue #10841 reproduction failed:', error);
		throw error;
	});
</script>
```

It feels like we could do better here by considering SPS from ISOBMFF.

youkaclub:
Fixes https://github.com/Vanilagy/mediabunny/issues/474

## Summary

- fall back to AVC SPS/VUI color metadata when ISO-BMFF has no `nclx`/`nclc` color box
- preserve container color metadata precedence and leave absent VUI fields unspecified
- pass the derived color space to `VideoDecoderConfig`

## Tests

- `npm test` (55 files, 514 tests)
- `npm run check`
- `npm run lint`
- `npm run build`
- verified the public issue reproduction reports `matrix: "bt470bg"` and `fullRange: true` for both the track and decoder config

### DIFF
--- a/src/codec-data.ts
+++ b/src/codec-data.ts
@@ -502,6 +502,8 @@ export type AvcSpsInfo = {
 	transferCharacteristics: number;
 	matrixCoefficients: number;
 	fullRangeFlag: number;
+	videoSignalTypePresentFlag: number;
+	colourDescriptionPresentFlag: number;
 	numReorderFrames: number;
 	maxDecFrameBuffering: number;
 };
@@ -657,6 +659,8 @@ export const parseAvcSps = (sps: Uint8Array): AvcSpsInfo | null => {
 		let transferCharacteristics = 2;
 		let matrixCoefficients = 2;
 		let fullRangeFlag = 0;
+		let videoSignalTypePresentFlag = 0;
+		let colourDescriptionPresentFlag = 0;
 		let pixelAspectRatio: Rational = { num: 1, den: 1 };
 
 		let numReorderFrames: number | null = null;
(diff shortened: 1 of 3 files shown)


Vanilagy:
Fix was insufficient, full fix in https://github.com/Vanilagy/mediabunny/commit/7d1eb5b063571073d11cda9a526aa323e5160ca0

Issue #80, opened 2025-08-29
Range iteration should be cleaned up if iterator throws


JonnyBurger:
The following warning appears if you let this code for for a few seconds:

```
A VideoFrame was garbage collected without being closed. Applications should call close() on frames when done with them to prevent stalls.
```

```ts
import {ALL_FORMATS, Input, UrlSource, VideoSampleSink} from 'mediabunny';

export const repro = async (src: string) => {
	const input = new Input({
		formats: ALL_FORMATS,
		source: new UrlSource(
			'https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4',
		),
	});

	const track = await input.getPrimaryVideoTrack();
	if (!track) {
		throw new Error(`No video track found for ${src}`);
	}

	const videoSampleSink = new VideoSampleSink(track);

	while (true) {
		const samples = videoSampleSink.samples();
		for (let i = 0; i < 20; i++) {
			const sample = await samples.next();
			if (sample.value) {
				sample.value.close();
			}
		}

		try {
			await samples.throw(new Error('Abort and decode'));
		} catch {}
	}
};
```

In this code, I am iterating over samples, but then later I realize I have read enough and I want to discard the iterator.
It seems like the samples which have not yet been `yield`-ed at the point where I throw are not being closed.

Version: 1.11.2

Vanilagy:
The proper way to terminate the iterator early is to use the `.return()` method, which is also the method that gets called by the browser when `break`ing out of an async iterator loop.

Not sure on the exact semantics of `throw()`, to be honest, but for now I say just don't use it.

This API will be reworked in the future anyway, so I'll just leave it as-is for now!

PR #484, opened 2026-09-04
Terminate sample iterators before throwing so late-decoded samples are closed

hijack2ee:
## Problem

In both sample generators in `src/media-sink.ts` (`mediaSamplesInRange` and `mediaSamplesAtTimestamps`), the `next()` error branches throw without setting `terminated`:

```ts
if (track.input._disposed) {
    closeSamples();
    throw new InputDisposedError();
} else if (hasOutOfBandError) {
    closeSamples();
    throw outOfBandError;
}
```

Per the language spec, `for await` does **not** call `return()` on an iterator whose `next()` threw — the iterator is simply abandoned. Since `terminated` (and `ended` in `mediaSamplesInRange`) stays `false`, the pump's decoder callback keeps pushing decoded samples into `sampleQueue` after the throw. No consumer will ever call `next()` again, `closeSamples()` already ran, and `return()` will never be called — those samples are orphaned until GC ("A VideoSample was garbage collected without first being closed"). This is the throw-path sibling of #80, which fixed cleanup for the `break`/`return()` path.

## Real-world impact

On iOS Safari, the hardware `VideoDecoder` can die mid-iteration under memory pressure (native error: `Decoding task did not complete`), which surfaces exactly through the out-of-band error path. Each orphaned `VideoSample` pins a GPU-backed `VideoFrame` until GC, so on a memory-pressured device the decoder failure *amplifies* the very pressure that caused it. We hit this in a multi-track video editor during heavy scrubbing, where it fed into further decoder failures and WebGL context losses.

Measured on-device (iPhone, iOS 18.7, six 1080p sources) with create/close accounting instrumentation on `VideoSample`: **3–14 samples orphaned per failing iterator per decoder-death event**, permanently (the created−closed residual never recovers). Note the `A VideoSample was garbage collected…` log is rate-limited to 1/s, so observed warning counts understate the leak.

## Fix

Set the termination flags before throwing, in the `InputDisposedError` and out-of-band error branches of both generators (`ended` too in `mediaSamplesInRange`, whose pump callback gates on it). Samples emitted by the decoder callback after the throw are then closed on arrival by the existing `terminated`/`ended` checks.

Verified on-device under the same reproduction: orphans dropped from 3–14 per iterator to **≤1**.

## Known residual

A sample materializing between the error being recorded and the consumer's `next()` running (i.e. already in flight inside the decoder wrapper during that microtask window) can still slip through — we consistently saw at most one per event. Closing that fully would need the pump to re-run `closeSamples()` once the consumer is known to be gone; this one-line-per-branch fix removes the unbounded part of the leak. Happy to extend the patch if you'd prefer that too.

`npx tsc --noEmit` and `eslint` pass on the changed file.

### DIFF
--- a/src/media-sink.ts
+++ b/src/media-sink.ts
@@ -601,11 +601,18 @@ export abstract class BaseMediaSampleSink<
 			async next() {
 				while (true) {
 					if (track.input._disposed) {
+						// Once next() throws, the consumer will never call return(), so terminate the
+						// iteration here - otherwise, the pump keeps queueing decoded samples that
+						// nothing can ever close.
+						terminated = true;
+						ended = true;
 						closeSamples();
 						throw new InputDisposedError();
 					} else if (terminated) {
 						return { value: undefined, done: true };
 					} else if (hasOutOfBandError) {
+						terminated = true;
+						ended = true;
 						closeSamples();
 						throw outOfBandError;
 					} else if (sampleQueue.length > 0) {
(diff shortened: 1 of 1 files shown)


Vanilagy (approved):
Thank you.

PR #489, opened 2026-09-08
Avoid unhandled rejections from failed prefetches

lideen:
## Use case

We use Mediabunny to play media through expiring S3 presigned URLs. When a required read fails, our player catches the error, obtains a fresh URL, and replaces the input while preserving the playhead and playback intent.

During this recovery flow, `ReadOrchestrator` also produces unhandled promise rejections from speculative reads, even when the application handles the required-read failure. A read served entirely from cache can trigger one of these rejections through its background prefetch.

## Change

- Handle the prefetch promise when the requested bytes are already cached.
- Stop rethrowing worker failures when no pending reads are waiting for data.

Pending reads still reject. Later reads retry missing bytes and reject if the source still fails, so the application can detect lost access and renew the URL.

## Tests

Two regression cases cover cached reads spanning one or two response chunks. A mocked HTTP boundary serves a prefix, then returns 403 for subsequent ranges. The tests check that cached reads succeed without unhandled rejections, required reads reject with 403, and a fresh source can read the missing bytes.

Both cases fail on current `main` and pass with this change.


Vanilagy:
Your problem is understandable, but I think simply swallowing the error is never a good idea. I now added a new `handleUnhandledError` callback option to most sources, including `UrlSource`. You'll be able to hook into that to be notified of any otherwise-unhandled errors. You can still choose to ignore them!

Let me know if this solves your problem adequately.

lideen:
> Your problem is understandable, but I think simply swallowing the error is never a good idea. I now added a new `handleUnhandledError` callback option to most sources, including `UrlSource`. You'll be able to hook into that to be notified of any otherwise-unhandled errors. You can still choose to ignore them!
> 
> Let me know if this solves your problem adequately.

This works, tyvm!

Issue #492, opened 2026-09-11
Expose full MP4/MOV presentation transforms


Vardiak:
Is there a way to read the full presentation transform of an MP4/MOV file?

`getRotation()` exposes rotation, but files can also contain scaling, translation, or reflection in their movie (`mvhd`) and track (`tkhd`) matrices.

It would be useful to access those transforms when drawing decoded frames manually, so we can apply them or detect files we cannot render correctly.

Would you consider exposing the matrices, or a combined transform, through the public API? Happy to contribute if this sounds useful.

Vanilagy:
Yes, important, agreed. I'm addressing this in https://github.com/Vanilagy/mediabunny/pull/510.

Vanilagy:
I added `getTransformationMatrix()` to `InputVideoTrack` now; this is the combined (multiplied) transformation (combining both movie and track matrices). There's also `getFlip()` now on video tracks, and flipping as a concept is now wholly supported as a first-party concept in Mediabunny. This means reading any video that contains any combination of rotation and flip works correctly in Mediabunny. That's all that Chrome supports anyway, so I don't think you need to fall back to `<video>` at all for this!

Translation and other transformations do still exist, and you can extract them with `getTransformationMatrix()`, but most players ignore these and extract just rotation and flip.

Let me know if this solves it for you!!

PR #493, opened 2026-09-11
Expose MP4 and MOV presentation matrices

Vardiak:
Addresses #492.

Adds `getTransformationMatrix()` to `Input` and `InputVideoTrack`, exposing the movie (`mvhd`) and track (`tkhd`) matrices. This lets callers inspect transforms beyond rotation, including mirroring, scaling, and translation, for custom rendering or detecting unsupported transforms.

The methods return copies of the stored matrices, or `null` when unavailable. Documentation explains their coordinate conventions and composition order. Pixel aspect ratio, cropping, and track sizing remain separate; existing rendering behavior is unchanged.

Includes tests for both header versions, initialization segments, and existing MP4, MOV, and WebM fixtures. All 393 Node tests, the build, type checks, and lint on changed files pass.

Written using GPT-6 Astra.


Vanilagy:
Thanks for the PR! I realized there's a good opportunity for feature expansion here, by having Mediabunny properly handled arbitrary transformation matrices as well as support flipping in addition to rotation natively. Therefore, this PR is superceded by https://github.com/Vanilagy/mediabunny/pull/510

PR #497, opened 2026-09-12
Fix NaN samples when upmixing stereo to quad or 5.1

meiiie:
Upmixing stereo to quad or 5.1 can write `NaN` instead of silence into the added channels at input chunk boundaries. The mixer reads past the stereo source view, then evaluates `undefined * 0`.

Return zero for the added channels before reading the source. Left/right mapping and interpolation are unchanged.

## Regression coverage

Ten tests exercise `AudioSampleSource` → float32 PCM WAV → `AudioSampleSink`: 4/6-channel output, one/two-frame inputs, uneven chunks, and simultaneous sample-rate conversion. They check finite PCM, exact silence, left/right samples, timestamps, metadata, and frame counts. All ten fail on baseline `cee57d1` and pass with this fix.

```sh
npm test -- --project node audio-resampling.test.ts
```

## Validation

Fork CI on Ubuntu with Node 22.23.2:

- [Node and Chrome tests](https://github.com/meiiie/mediabunny/actions/runs/34687815677): **562 passed across 57 files**, using `xvfb-run npm test`.
- [TypeScript and ESLint](https://github.com/meiiie/mediabunny/actions/runs/34687815673): passed.
- [Library, documentation, and examples build](https://github.com/meiiie/mediabunny/actions/runs/34687815706): passed.

These runs use [862aaf0](https://github.com/meiiie/mediabunny/commit/862aaf07d5fa80272a9cda28b9e650a5bf1287de), which differs from the PR head only in the three workflow push triggers and an EOF newline. Source, tests, dependencies, and CI commands are identical. **Upstream PR workflows are awaiting maintainer approval.**

<details>
<summary>Additional local validation</summary>

- Windows/Chrome: 161 passed, two negative-timestamp MPEG-TS tests timed out. Both timeouts reproduced on the unmodified baseline.
- The full Windows Node run encountered fixture-path and worker errors. WSL typecheck passed, but ESLint exhausted the local heap; remaining WSL checks were stopped after fork CI passed.

</details>

AI disclosure: Codex assisted with the investigation, implementation, tests, and PR description.


Vanilagy (approved):

Vanilagy:
Great catch, thanks for the fix. I removed the test file since it was somewhat specific and this was the only remaining bug in the audio mixing code, now everything is correct I think!

PR #499, opened 2026-09-13
Cancel CustomSource streams when disposing inputs

meiiie:
Disposing an input backed by a `CustomSource` rejects pending reads, but leaves stream readers waiting for data and does not cancel their underlying streams. An asynchronous `read` callback can also return a stream after disposal, which is then abandoned.

Track active readers and cancel them on disposal, cancel streams returned after disposal, and release reader locks when reads finish or fail. The existing disposal callback and `InputDisposedError` behavior are preserved. Regression coverage is added to the existing source lifetime tests for pending and late streams, cancellation failures, and reader cleanup.

Upstream CI on `860e4f5`: [Test](https://github.com/Vanilagy/mediabunny/actions/runs/34760869770) passes all 558 tests; [Lint](https://github.com/Vanilagy/mediabunny/actions/runs/34760869766) and [Build check](https://github.com/Vanilagy/mediabunny/actions/runs/34760869679) pass.

Additional validation on Node 22.23.2 in WSL/Linux:

- The six new regression cases fail on `28ed500`; all nine source lifetime tests pass with the fix. The wider input/source/MP4/WAV selection passes 22/22 tests.
- `npm run check`, ESLint on the changed files (with a 4 GB heap), and `npm run pre-test` pass.
- A local HTTP reproduction leaves a delayed response open after disposal on the baseline; with the fix, the server observes it closing within the 200 ms observation window, without an additional explicit abort.

AI assistance: implemented and tested with Codex.


Vanilagy (approved):
Looks good to me

PR #505, opened 2026-09-15
feat(mp4): derive track bitrate metadata automatically

ntvsx193:
## Summary

Write MP4 bitrate metadata automatically from encoded samples or the selected encoder configuration, without requiring callers to supply track metadata.

- For non-fragmented MP4 (`fastStart: false`, `in-memory`, and `reserve`), measure the average bitrate over the samples' presentation span and the maximum number of payload bits in any fixed one-second window. Use presentation timestamps at track timescale precision and payload sizes after format conversion.
- For fragmented MP4, where `moov` is written before the complete stream is available, use the selected encoder's nominal bitrate after `onEncoderConfig`. Quantizer encoding and packet sources with no known nominal bitrate leave video `btrt` absent and audio `esds` bitrate fields at zero.
- Write video `btrt` and the existing audio `esds` maximum/average fields. Clamp values to unsigned 32-bit range and account for `btrt` when reserving space for `moov`.
- Remove the previously proposed public `metadata.bitrate` fields and validation.

The maximum uses a fixed one-second window even for clips shorter than one second; the average uses the actual presentation span. Regular MP4 can measure bitrate regardless of encoder rate-control mode.

## Validation

- 279 Node tests passed across 29 test files, including byte-level MP4 assertions, all regular `fastStart` modes, B-frame order, timestamp shifts and gaps, exact one-second boundaries, short/zero-duration clips, ADTS stripping, uint32 bounds, and encoder configuration/quantizer fallback integration.
- Regression tests failed before the implementation and passed afterward.
- `npm run check`, `npm run lint`, `npm run build`, and `git diff --check` passed.
- Four Node test files (`hls-input`, `prores`, `url-source`, and `server-extension`) were excluded from this run because the unchanged upstream baseline already failed on unavailable network fixtures/local listening sockets in this environment. Browser tests were not run.


Vanilagy (changes requested):
I think we can do better than this.

Having to pass the bitrate as metadata is additional work for the user and is also redundant work since Mediabunny is typically already aware of how the encoder was configured, and thus knows the nominal bitrate. Plus, the fields *are* defined as:

```
maxBitrate	gives	the	maximum	rate	in	bits/second	over	any	window	of	one	second.	
avgBitrate	gives	the	average	rate	in	bits/second	over	the	entire	presentation.	
```

For the variants of ISOBMFF writing where the moov box can be written/modified at the end, the muxer can literally just measure these values exactly.

ntvsx193:
How should we handle fragmented/streaming MP4, where the `moov` box is written before the actual bitrate can be measured? Would it make sense to use the target bitrate from the encoder configuration as a fallback in that case?

Vanilagy:
Yes! For those cases where we cannot patch the `moov` box, we can fall back to the nominal values extracted from the `VideoDecoderConfig` and `AudioDecoderConfig` that the muxer receives. If we encode in quantizer mode, we can't write anything since we don't know the bitrate ahead of time.

Vanilagy (approved):
Modified the logic a little. Behavior is mostly the same. You can also now specify the bitrate in the track metadata, but it will be overwritten if a more accurate value can be determined at mux time.

PR #507, opened 2026-09-15
feat(encoding): support frame rate and config-only video checks

ntvsx193:
## Summary

Allow video capability checks to include the intended frame rate and optionally query native configuration support without creating an encoder. For example, a 1080×1920, 30 FPS export can now check its target preset explicitly instead of relying on the default dimensions.

- Add optional finite, positive `framerate` to `canEncodeVideo`.
- Add `checkMode: 'auto' | 'config-only'`, defaulting to the existing `auto` behavior.
- In `config-only`, query `VideoEncoder.isConfigSupported` directly, bypass custom encoders, trial encoding, and the shared cache. Return false when WebCodecs is missing or AVC/HEVC dimensions are odd; preserve native query rejections.
- Reuse `canEncodeVideo` options in `getEncodableVideoCodecs` and `getFirstEncodableVideoCodec`.

The implementation preserves the current Quality API and tries its native configuration candidates in preference order. A positive configuration check does not guarantee a successful full encode.

## Validation

- 11 Node tests passed, covering FPS forwarding, invalid options, absent WebCodecs, odd dimensions, custom encoder/cache bypass, candidate fallback, native query errors, and helper forwarding.
- Regression tests failed before implementation and passed afterward.
- `npm run check`, `npm run lint`, and `npm run build` passed.
- Browser boundaries were stubbed in the Node tests; browser acceptance was not run. The full unchanged-upstream Node baseline is limited by unavailable network fixtures and listening sockets in the restricted environment.


Vanilagy:
Don't see a point in `config-only` if it then just ends up calling `isConfigSupported` anyway, which is exactly what the current code does.

PR #508, opened 2026-09-15
feat(encoding): expose native encoder support and error observers

ntvsx193:
## Summary

Expose optional video/audio encoder observers so callers can diagnose native support checks and encoding failures without changing encoder selection or error handling.

- Add `onEncoderSupport(config, event)` for existing native queries, with `started`, `supported`, `unsupported`, and `error` events. Each notification receives a detached configuration snapshot, including nested codec options.
- Add `onEncoderError(error, stage)` for native `create`, `configure`, `encode`, `flush`, and asynchronous `error` failures, including the alpha encoder. Native operation errors retain their original identity and propagation.
- Export `EncoderSupportEvent` and `EncoderErrorStage` and validate both callbacks. Synchronous exceptions thrown by observers are ignored.

Keep the current native candidate fallback behavior: a rejected support query is reported with its original error before the existing selection flow continues; it may ultimately produce the existing aggregate unsupported-configuration error. Custom encoder selection is unchanged. The callbacks provide hooks only; this change does not send telemetry or implement a retry policy.

## Validation

- 57 Node tests passed across encoder observers, output, and empty media. Observer cases cover validation, nested snapshot isolation, observer exceptions, support outcomes, error identity, and alpha encoder failures.
- Regression tests failed before implementation and passed afterward.
- `npm run check`, `npm run lint`, and `npm run build` passed, including API extraction and public docblock validation.
- WebCodecs boundaries were stubbed in Node; browser acceptance was not run. The full unchanged-upstream Node baseline is limited by unavailable network fixtures and listening sockets in the restricted environment.


Vanilagy:
Adds a bit too much complexity for my liking, so I won't merge this in right now. Also, providing an `onEncoderError` callback is misguided: it confuses the user as to if they actually need to use that callback to be notified of encoder errors. Encoder errors are forwarded to the user anyway within the sync logic of the code, so that callback is redundant and confusing.

It would help (and this is a general point for your PRs) if you actually described your *concrete* problem. These PRs tell me nothing about what difficulties you're actually having, they just jump straight to the solution.

ntvsx193:
Our use case is a video editor running on Android, iOS, and desktop browsers. We have several candidate codecs for encoding, and their availability depends on the device, browser, and requested encoding settings.

When the editor starts, we need to identify which codecs the device reports as supported and choose an initial encoding configuration. The `config-only` mode in #507 lets us query native support without creating encoders or encoding test frames for every candidate, keeping this startup check lightweight.

However, a supported configuration does not guarantee that encoding will succeed at runtime. This is where the callbacks in #508 complement that check:

- `onEncoderSupport` exposes the configuration actually checked during encoder initialization and the result of that check.
- `onEncoderError` exposes the original error and the stage where it occurred, such as configuration, encoding, or flushing.

Together, these APIs let us make an initial codec choice at startup, diagnose device-specific failures during export, and decide whether to retry with another codec or different settings in our application.

PR #509, opened 2026-09-16
Fix HLS AES-128 IV parsing

Pheo:
## Problem
Parser was advancing every half-byte (4-bits / 1 hex digit) instead of 1 byte (8-bits / 2 hex digits). This produces wrong IV and results in corrupt first block of each segment.

relevant spec: https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.4.4

Given:
```m3u8
#EXT-X-KEY:METHOD=AES-128,URI="key.bin",IV=0x00112233445566778899aabbccddeeff
```
The expected IV is:
```text
00 11 22 33 44 55 66 77 88 99 aa bb cc dd ee ff
```
But the parser instead produces:
```text
00 01 11 12 22 23 33 34 44 45 55 56 66 67 77 78
```

## Repro

Standalone AES-128 Explicit-IV HLS Repro: [aes128-iv-hls-repro.tar.gz](https://github.com/user-attachments/files/32276965/aes128-iv-hls-repro.tar.gz)

```
tar -xzf aes128-iv-hls-repro.tar.gz
cd aes128-iv-hls-repro
node server.mjs
```

Broken in mediabunny@1.56.3:
```
const mediabunny = await import('https://cdn.jsdelivr.net/npm/mediabunny@1.56.3/dist/bundles/mediabunny.mjs');
const input = new mediabunny.Input({
    source: new mediabunny.UrlSource('http://127.0.0.1:8080/playlist.m3u8'), // aes128-iv-hls-repro
    formats: mediabunny.ALL_FORMATS,
});
await input.getPrimaryAudioTrack(); // throws UnsupportedInputFormatError
```

With our patch:
```
const mediabunny = await import('<fixed-parser.mjs>');
const input = new mediabunny.Input({
    source: new mediabunny.UrlSource('http://127.0.0.1:8080/playlist.m3u8'), // aes128-iv-hls-repro
    formats: mediabunny.ALL_FORMATS,
});
await input.getPrimaryAudioTrack(); // returns InputAudioTrack
```

## Misc
Manual decryption of `segment.aac`:
```
openssl enc -d -aes-128-cbc \
    -K "$(xxd -p -c 256 key.bin)" \
    -iv 00112233445566778899aabbccddeeff \
    -in segment.aac \
    -out decrypted.aac
```


Vanilagy (approved):
Good catch, thanks. However, a 32-character trim was still needed. That was the point of the negative index read in the first place! 

Also `2 * i` because `i * 2` looks disgusting

PR #513, opened 2026-09-17
Ignore Content-Range/Length on content-coded 206 responses

### LINKED ISSUE #512: UrlSource can get the wrong file size for 206 + Content-Encoding
## `UrlSource` can get the wrong file size for `206` + `Content-Encoding`

**Affects:** `UrlSource` in `src/source.ts`

I'm hitting this:

```text
Incorrect M3U8 file; #EXT-X-STREAM-INF must be followed by a URI line.
```

`UrlSource` sometimes thinks the file is much smaller than it really is, so the reader ends up parsing a truncated version of it.

If a server responds to a range request with both:

* `206 Partial Content`
* `Content-Encoding: gzip` (or another encoding)

then the `Content-Range` total is for the **encoded/compressed bytes**.

But `fetch`/`undici` gives us the **decoded** body.

So we end up doing something roughly like:

```text
Content-Range: bytes 0-99/1556
                         ^^^^
                    compressed size

decoded body: 12818 bytes
```

`UrlSource` takes `1556` as the file size, even though the reader is actually working with the 12818 decoded bytes.

That means we only read the first 1556 decoded bytes and can cut the M3U8 right in the middle of a tag.

This is what eventually turns into:

```text
#EXT-X-STREAM-INF must be followed by a URI line
```

For Apple's public test stream, I saw exactly this:

```text
HTTP/1.1 206 Partial Content
Content-Encoding: gzip
Content-Range: bytes 0-99/1556
```

The actual decoded file is 12818 bytes.

Whether you get it depends on the `Accept-Encoding` sent with the range request, measured against that URL:

```
Range: bytes=0- (Node fetch default)       200 | identity | content-length=12818
Range: bytes=0- + Accept-Encoding: gzip    206 | gzip     | content-range: bytes 0-1555/1556
no Range (Node fetch default)              200 | gzip     | content-length=1556
```

The decoded body is 12818 bytes in all three. Browsers always send `Accept-Encoding`, so the middle row is the normal case there; Node's defaults happen to avoid it.

```mermaid
sequenceDiagram
    participant R as Reader
    participant U as UrlSource
    participant S as CDN

    R->>U: requestEntireFile()
    U->>S: GET master.m3u8 (Range: bytes=0-)
    S-->>U: 206 + Content-Encoding: gzip<br/>Content-Range: bytes 0-99/1556
    Note over U: status is 206, so Content-Encoding<br/>is never checked
    U->>U: fileSize = 1556 (compressed)
    Note over U: fetch already decoded the body<br/>12818 bytes are really there
    U-->>R: slice claiming 1556 bytes
    R->>R: readAllLines(slice, 1556)
    R--xR: cut mid-tag
```

There is already a check for `Content-Encoding` in `src/source.ts`:

```ts
&& (response.status === 206 || (response.type === 'basic' && !response.headers.has('Content-Encoding')))
```

The problem is the `206` part.

If the response is `206`, the condition is already true, so we never care whether `Content-Encoding` exists.

There's also a comment saying:

> Content-Encoding is basically never used for range responses...

Unfortunately, that doesn't seem to be true in practice. The Apple CDN is doing exactly that.

(Both the comment and the condition come from `450c5def`, "Fix incorrect use of Content-Length when the server compresses the content (fixes #487)" — so this looks like the same issue reappearing through the `206` path.)

### Repro

This is enough to reproduce it:

```js
import { createServer } from 'node:http';
import { gzipSync } from 'node:zlib';
import { UrlSource } from './src/source.js';
import { Reader } from './src/reader.js';

const body = Buffer.from(
  '#EXTM3U\n' +
  '#EXT-X-STREAM-INF:BANDWIDTH=1\nv.m3u8\n'.repeat(120)
);

const gz = gzipSync(body);

const server = createServer((_req, res) => {
  res.writeHead(206, {
    'Content-Encoding': 'gzip',
    'Content-Range': `bytes 0-${gz.length - 1}/${gz.length}`,
    'Content-Length': gz.length,
  });

  res.end(gz);
});

await new Promise(r => server.listen(0, r));

const src = new UrlSource(
  `http://127.0.0.1:${server.address().port}/master.m3u8`
);

const slice = await new Reader(src).requestEntireFile();

console.log(
  'truth:',
  body.length,
  'reported:',
  slice.length
);

server.close();
```

For me this gives something like:

```text
truth: 5010 reported: 245
```

So `reported` is the gzip size instead of the decoded size.

### Fix

I think the check should just reject `Content-Encoding` before looking at the status:

```diff
- // Content-Range/Length fields are meaningless if Content-Encoding is present. Content-Encoding is
- // basically never used for range responses (since the encoding runs *before* the slicing), so we're set
- // in that case.
- && (response.status === 206 || (response.type === 'basic' && !response.headers.has('Content-Encoding')))
+ // Content-Range/Length refer to the encoded representation, while the reader works with decoded bytes.
+ // This can happen for 206 responses too.
+ && !response.headers.has('Content-Encoding')
+ && (response.status === 206 || response.type === 'basic')
```

That makes the existing fallback kick in and use the number of decoded bytes actually received instead.

I tested this with both gzip `200` and gzip `206` responses and the size comes out correctly after the change.

### Regression test

There's already a test:

`UrlSource reads the full decoded body of a compressed response`

It currently uses `200`.

Changing it to a `206` response with a `Content-Range` based on the compressed size reproduces the bug:

```ts
res.writeHead(206, {
  'Content-Type': 'text/plain',
  'Content-Encoding': 'br',
  'Content-Range': `bytes 0-${compressed.byteLength - 1}/${compressed.byteLength}`,
  'Content-Length': compressed.byteLength,
  'Accept-Ranges': 'bytes',
});
```

The existing assertion is enough:

```ts
slice.length === content.byteLength
```

Before the fix:

```text
expected 33 to be 4864
```

After the fix, it passes.

### How I found it

The `Advanced Apple HLS` test was failing about 1/3 of the time when running the full suite, but I couldn't reproduce it reliably in isolation.

I added some logging around the parse failure and eventually got:

```text
sliceLen=1556 bytes=3624 lines=14 truth=12818
tail="…RESOLUTION=960x540,FRAME-RATE=60.000,CLOSED-CAPTIONS="
```

The `1556` matching the compressed `Content-Length` was the giveaway.

After changing the `206` handling, I ran the full suite 11 times with the instrumentation still enabled and didn't see the truncation again.

So I think the issue is simply that we're trusting `Content-Range`/`Content-Length` even though `undici` has already decoded the response body.

ahmedrowaihi:
Fixes #512.


Vanilagy:
Merging one of your PRs for once!! 😂 ❤️

PR #514, opened 2026-09-17
Add BPM to ID3 metadata

Snazzah:
Taken from [`TBPM`](https://web.archive.org/web/20240827101950/https://id3.org/id3v2.3.0#:~:text=is%20taken%20from.-,TBPM,-The%20%27BPM%27%20frame) since its read in a few things.


Vanilagy:
Good idea! If this has an equivalent in Vorbis-style metadata used in FLAC/Ogg, and for ISOBMFF metadata boxes, we should support it there too.

Vanilagy (approved):
Thanks, clean. Renamed to `beatsPerMinute` since I avoid abbreviations in Mediabunny's API. I considered upgrading it to allow fractional BPMs as well, but most of the formats can do just int, so I left it at that.

PR #516, opened 2026-09-18
Add HLS Variable Substitution + #EXT-X-DEFINE tag

Pheo:
## What?
This PR implements:
- Variable Substitution: https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.3
- EXT-X-DEFINE: https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis#section-4.4.2.3

HLS Variable Substitution is a powerful feature that enables a lot of use-cases. For example, it can be used to insert auth tokens into segment fetch requests.

Given a playlist requested as:
```
https://example.com/media.m3u8?token=123
```

Containing:
```
#EXTM3U
#EXT-X-DEFINE:QUERYPARAM="token"
#EXT-X-TARGETDURATION:6

#EXTINF:6,
https://cdn.example.com/segment-1.ts?auth={$token}
```

The resulting segment request is:
```
https://cdn.example.com/segment-1.ts?auth=123
```


Pheo:
Switching back to draft for a bit. I'm missing tests in `test/node/hls-input.test.ts`
EDIT: added tests. looks like CI failing on unrelated flakey tests?

Vanilagy:
Oh you work at Clipchamp!

src/hls/hls-misc.ts
```diff
@@ -25,40 +28,167 @@ export const TAG_I_FRAMES_ONLY = '#EXT-X-I-FRAMES-ONLY';
...
+// variable name `token_1` is referenced as `{$token_1}`
+const VARIABLE_NAME_PATTERN = /^[A-Za-z0-9_-]+$/;
+const VARIABLE_REFERENCE_PATTERN = /\{\$([A-Za-z0-9_-]+)\}/g;
+
+export class HlsPlaylistVariables {
+	_variables = new Map<string, string>();
+
+	constructor(
+		readonly playlistPath: string,
+		readonly importedVariables: ReadonlyMap<string, string> | null,
```
Pheo: Could make this `HlsPlaylistVariables` instead of passing around a `Map<string, string>`.
```suggestion
		readonly importedVariables: HlsPlaylistVariables;
```

Then we change `masterPlaylistVariables` in `his-demuxer.ts` to be of type `HlsPlaylistVariables`

PR #517, opened 2026-09-18
Fix G.711 s16 conversion and ulaw/alaw codec recognition

cyanyux:
## Problem

G.711 (µ-law/A-law) audio is broken in three ways:

1. **Wrong amplitude and polarity.** `fromUlaw`/`fromAlaw` return values in the 14/13-bit companding domain (±8031 / ±4032), but `PcmAudioDecoderWrapper` writes them verbatim via `setInt16`. Decoded µ-law is ~12 dB too quiet; A-law is ~18 dB too quiet — and additionally polarity-inverted, since the sign convention is flipped relative to ITU-T G.711 (a set sign bit in the codeword means *positive*).
2. **`canDecode()` false negative.** `InputAudioTrack.canDecode` only treats `pcm-*` codecs as software-decodable, so `ulaw`/`alaw` tracks report undecodable in environments without `AudioDecoder` — even though `PCM_AUDIO_CODECS`, the demuxers, `canDecodeAudio()`, and `PcmAudioDecoderWrapper` all already support them.
3. **Symmetric encoder bug.** `toUlaw`/`toAlaw` operate on the companding domain but `media-source.ts` feeds them full-scale s16, so any sample above quarter-scale clamps to the maximum codeword. `toAlaw` also shares the inverted sign convention.

## Fix

- `fromUlaw` output ×4 and `fromAlaw` output ×8 with corrected polarity. Decoding is bit-exact against the ITU-T G.711 tables for all 256 byte values (e.g. µ-law `0x00→-32124`, `0x80→32124`; A-law `0xD5→8`, `0xAA→32256`).
- `toUlaw`/`toAlaw` shift s16 input into the companding domain (`>>2`/`>>3`), `toAlaw` sets the sign bit for positive samples, and negative inputs use `-n - 1` to match the canonical reference encoder exactly.
- `canDecode` uses `PCM_AUDIO_CODECS.includes(codec)` — the same check as `canDecodeAudio`, so `ulaw`/`alaw` are handled by the software decoder without requiring `AudioDecoder`.

## Tests

New `test/node/g711.test.ts`: ITU-T vectors for decode, encode pins, encode/decode round-trips, and end-to-end µ-law/A-law WAV decoding through `Input` + `AudioSampleSink` (exercising `canDecode` under Node, where no `AudioDecoder` exists). WAV-path assertions allow ±1 LSB since `copyTo('s16')` rounds through the normalized f32 intermediate.

`npx vitest --run node/` — all tests pass; 6/7 new tests fail on `main` (the encode pins and WAV decode cases), and the round-trip fails on the old polarity/scale.


Vanilagy (approved):
Good catch; made a bunch more fixes to PCM while I was at it. Reorganized the tests.

PR #518, opened 2026-09-18
Abort pending UrlSource fetches on dispose

cyanyux:
## Problem

`UrlSource._dispose` disposes the read orchestrator but never signals the network layer, so pending HTTP requests outlive the source:

- **Ranged mode:** each worker fetch uses a per-iteration `AbortController` that is only aborted once a pending `reader.read()` resolves. If the server stalls mid-response, disposing leaves the fetch — and the TCP connection — alive indefinitely.
- **Sequential mode:** the resume fetch inside the wrapped stream's `pull` carries no signal at all, so a stalled reconnect is never released.
- **Unhandled rejections:** `ReadableStreamSource._dispose` calls `reader.cancel()` without handling the returned promise, and the sequential transition does the same with `response.body.cancel()` — both surface as unhandled rejections when the cancel fails (the `CustomSource` fix in 51fd5ac addressed the same pattern).

## Fix

- Add a source-level `_disposeController`, aborted in `_dispose`. Each worker fetch wires it to its per-iteration controller via a `once` abort listener (ES2021-safe; `AbortSignal.any` would require Safari 17.4+/FF 124+), and the sequential-mode resume fetch uses the dispose signal directly. The retry predicate also consults the controller so an aborted request is never retried.
- Attach `.catch(() => {})` to both dropped `cancel()` calls.

## Tests

`test/node/url-source.test.ts` gains two cases against local `http` servers: a range-capable response that stalls mid-body, and a rangeless server whose resume request stalls — both assert the server observes the connection closing after dispose. `test/node/source-lifetime.test.ts` gains a case asserting a rejecting `cancel()` produces no unhandled rejection.

All three fail on `main` (the ranged test additionally surfaces an unhandled socket error there) and pass with this change.


Vanilagy:
I'm sorry, did you actually run into this or are you just fixing stuff just cause?

.dispose() *does* call the AbortController as soon as the worker enters the next iteration. "If the server stalls" seems arbitrary to me. And have you ever in your life seen `.cancel()` fail? 😂

cyanyux:
Yes, we did run into it — though to be precise, we found it auditing disposal behavior in our app and then confirmed it with real-socket tests, rather than hitting a user-facing incident first. Context: we run a video wall that tears down decode pipelines constantly (tile recycling, quality swaps, reconnects, unmounts), often while a worker is parked on a read for bytes that never arrive. On the installed dependency the parked read rejects with InputDisposedError, so the app thinks teardown finished, but the server-side socket stays open. The tests in this PR reproduce that on main with real HTTP servers — a range response that stalls mid-body, and a sequential-mode resume that stalls — and on the patched build Chrome reports net::ERR_ABORTED when a player unmounts mid-open.

On the mechanism: you're right that worker.aborted gets set and the abort fires at the top of the loop. But that only works once the loop iterates, and both awaits can park indefinitely — `await retriedFetch(...)` (source.ts:947) when the server never sends headers, and `await reader.read()` (1062) on a stalled body. If read() never resolves, `worker.aborted` is never observed. There's also no signal at all on the sequential-mode resume fetch at 1146 — `mergeRequestInit` only sets the Range header — so if the wrapped stream is cancelled mid-resume, `currentReader.cancel()` hits the dead reader and the in-flight fetch is unreachable.

As for `.cancel()`: a stream that's already errored rejects cancel(), and our sequential-retry teardown surfaced a real unhandled rejection on main. Worth noting the same file already swallows cancel rejections at 506, 568, and 1626 (including the CustomSource fix in 51fd5ac) — this just covers the two sites that were missed (1016 and 1935). The wrapped-stream cancel at 1205 makes it slightly worse: a rejecting inner cancel propagates to the outer reader.cancel() at 1935.

Happy to trim this to just the parts you consider in scope.

Vanilagy:
Fair points! I implemented it myself in the attached commit, your cases should be covered now. Requests are aborted immediately on dispose, and cancels can't throw no matter what (their unhandled rejections are useless info anyway).

PR #519, opened 2026-09-20
Fix rounding errors at audio resampling boundaries

meiiie:
Floating-point error in the resampling end position can make `ceil` include an extra output frame. For example, transforming 17 frames from 16 to 48 kHz produces 52 frames instead of 51; channel-only transforms can also gain a frame.

At an output buffer boundary, that extra frame can flush the buffer before the next input sample contributes its interpolation. With 16→48 kHz PCM starting at 0.1 seconds and chunks of 14439, 65561 and 16000 frames, a constant 0.75 signal drops to 0.5 and 0.25 immediately before the five-second boundary. Passing the same input as one sample avoids the dip.

Apply the existing `roundIfAlmostInteger` helper before rounding the end position up. Regression tests in the existing PCM suite cover frame counts, channel-only transforms and interpolation across buffers, with fractional-endpoint controls to ensure those still round up. This retains the existing linear interpolation and adds no API or dependencies.

Upstream CI on `2d730b6`: [Test](https://github.com/Vanilagy/mediabunny/actions/runs/35521188396) passes all 612 tests in 58 files; [Lint](https://github.com/Vanilagy/mediabunny/actions/runs/35521188333) and [Build check](https://github.com/Vanilagy/mediabunny/actions/runs/35521188397) pass.

Additional validation on Node 22.23.2 in WSL/Linux:

- On main `bad03f2`, four regression cases fail and the other 18 PCM tests pass; with the fix, all 22 pass.
- `npm run check`, full ESLint (4 GB heap), and `npm run docs:build` pass.

<details>
<summary>Local full-suite limitations</summary>

The Node suite finished with 425 passes and 14 failures (HLS timeouts, AV1 codec-open errors and a ProRes timeout). Running the two affected files on production baseline `bad03f2` reproduced 12 of those failures; the other two cases passed when rerun on the patched code. This is not a green full local run. Browser tests were not run locally; upstream CI provides the full Node/Chrome check. Local dependencies were installed with `SKIP_FFMPEG=true`, skipping the optional FFmpeg CLI download.

</details>

AI assistance: implemented and tested with Codex.


Vanilagy:
This is indeed a bug, but your fix was too narrow

meiiie:
> This is indeed a bug, but your fix was too narrow

Thank you for pointing that out, I will look into it more thoroughly. As a newcomer, I’m really glad to be able to contribute to the mediabunny community.

Issue #345, opened 2026-04-08
Port over size optimizations from my AAC encoder extension?


valadaptive:
I started working on a project that needed AAC encoding support before the `@mediabunny/aac-encoder` package was created. As such, I created [my own AAC encoder package](https://github.com/ntsc-rs/ntsc-rs-web/tree/63fd1f11a1e1e967737bd16b972cb1c4955ae218/aac-codec). It's similar to `@mediabunny/aac-encoder`, compiling FFmpeg to WebAssembly with all features disabled except for the AAC encoder, but I also included a number of aggressive size-optimization patches.

The size optimizations are listed in the README, but I'll list them here as well. They consist of removing features that are unused at runtime, but cannot be eliminated by the compiler:
- I removed option parsing entirely, and manually set the relevant struct fields instead. The option parsing code is extremely heavy, and contains (among other things) large tables for parsing dates and X11 color names, as well as human-readable *description text* for every option.
- I patched out a couple of code paths where video frame info was being printed or inspected. This was pulling in another large table of video pixel formats (and their human-readable string names).
- I removed double-precision and fixed-point MDCT support. The AAC encoder never uses those, but libavutil's MDCT uses dynamic dispatch, so they cannot be statically eliminated.
- I removed the two-loop encoding method, and chose to always use the "fast" method. This is a functional deviation from FFmpeg's defaults, but the documentation says:
  > Uses a cheaper version of twoloop algorithm that doesn’t try to do as many clever adjustments. Worse with low bitrates (less than 64kbps), but is better and much faster at higher bitrates.

  My application always encodes at 128kbps, so it's not a problem for me. However, I'd imagine you might want to leave the two-loop algorithm enabled since you're distributing a library and allowing the consumers to control the bitrate.

These patches bring the WebAssembly bundle size down to 226KB uncompressed, or ~100KB zipped. If you want, feel free to port over any of my code.

---

[...]

PR #523, opened 2026-09-21
Reduce AAC encoder size by omitting unused transforms

meiiie:
Port [valadaptive's float-only transform patch](https://github.com/ntsc-rs/ntsc-rs-web/blob/63fd1f11a1e1e967737bd16b972cb1c4955ae218/aac-codec/patches/0002-tx-float-only.patch) to omit unused double-precision and fixed-point transforms from the AAC build. Keeps the default two-loop encoder and public API.

With the same FFmpeg revision and compiler, the minified ESM bundle shrinks from **254,104 to 211,545 bytes with gzip level 9 (-16.75%)**. The README pins the source/compiler versions and documents the patch. The previous artifact's exact build versions were not recorded.

Validation:

- Packet bytes, configuration and timestamps match in 108 configurations against both the unpatched rebuild and the existing artifact.
- Existing AAC tests pass on Node 22; browser Worker roundtrip and independent FFmpeg decoding pass.
- `npm run check`, `npm run lint`, `npm run docs:build`.

Refs #345.

AI assistance: implementation and validation with OpenAI Codex.


Vanilagy:
Eh, for the 40k gzipped bytes I'll pass on this for now. The patch would have to hypothetically change if FFmpeg's code were to change here.

But, can reconsider in the future.

